PR #1, opened 2024-04-04
Initial notes

lewisjared:
# Description

Some notes from reading through the codebase about some initial code cleanup that would be useful, including some questions for @prayner. 

The biggest hole for me is understanding where the files are downloaded. I'm particularly interested in understanding how the data flows between these repos.

I can also move these notes into a GDoc if that is a better medium (I was offline when writing these notes).

CC @aethr

### DIFF
--- a/notes/initial-cleanup.md
+++ b/notes/initial-cleanup.md
@@ -1,0 +1,53 @@
+# Notes
+After a read through of the codebase, I've jotted down some thought about the steps to clean up this repository.
+Broadly it is easy enough to follow where things live, but the goal is to improve the consistency of this repo
+(and other OM repos). This should make it easier to work as a team and produce extensible software in the long-run.
+
+The goal would be to make no science changes as part of this refactor, i.e. the output data should not change.
+
+This clean-up might not be the #1 priority, but should go on the roadmap. Once we agree on the steps involved,
+we can create a set of GH issues to track the process.
+
+## Steps:
+* Add an integration test
+    * Might need to use an environment variable to run in "reduced complexity mode"
+    * Run through from start to finish and verify that we get the same result.
+* Run ruff over the codebase. This will coalesce on a common variable/function naming conventions
+* Refactor code into a package. Working name `om_prior`
+    * Split out the processing steps from the utilities
+    * Define an interface that all layers must implement
+    * Move each of the layer scripts into a common place
+* Move inputs,intermediates,outputs into a data folder
+* Migrate to the CR copier template to define the repository structure and workflows. This will add linting, CI, versioning etc
+    * This will result in some changes in the development workflow so there might be some training needed
+    * We should discuss what steps work for the rest of the team. There are part of the template that we can pick and choose
+    * We've documented the technical decisions that we have made which we can share
+* Run CI periodically to double-check that the processing is always ready
+* Add licenseheaders to automate the generation of the license headers in all source files
+
+## Questions for Peter
+* How does the downloading of inputs work? I see that data is fetched from `https://prior.openmethane.org`. How does data get there?
+* What do you do with the outputs? Are they uploaded to `https://prior.openmethane.org`?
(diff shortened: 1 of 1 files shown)


notes/initial-cleanup.md
```diff
@@ -0,0 +1,59 @@
...
+    * This will add linting, CI, versioning etc
+    * There will be some changes in the development workflow so there might be some training needed
+    * We should discuss what steps work for the rest of the team. There are part of the template that we can pick and choose
+    * We've documented the technical decisions that we have made which we can share
+* Run CI periodically to double-check that the processing is always ready
+* Add licenseheaders to automate the generation of the license headers in all source files
+
+## Questions for Peter
+* How does the downloading of inputs work? I see that data is fetched from `https://prior.openmethane.org`. How does data get there?
+* What do you do with the outputs? Are they uploaded to `https://prior.openmethane.org`?
```
aethr, line 36: Gerard showed me yesterday that prior.openmethane.org is a R2 bucket (CloudFlare version of S3). His understanding is that the prior doesn't change very frequently unless new data or methodology becomes available. This can be computed based on (fairly) static inputs and placed in an easily accessible location for reference by the model.

There is some utility to this data and the methods for generating it being publicly available.

It probably makes sense for the prior to be "productionised" but possibly not automated. @prayner what are your thoughts?

lewisjared: At a glance it looks like the data in the R2 bucket is only the semi-static data that is updated semi-annually. I'm fine if these data are updated on an ad hoc basis rather than being checked monthly and automatically used. The step of "productionising" those semi-static data might be as simple as documentation (public or private).
 
The [GFAS data](https://github.com/openmethane/openmethane-prior/blob/main/omGFASEmis.py) is downloaded during the processing of its layer and will be updated at least monthly.

@aethr When you say "prior" are you referring to this repository as a whole or just the data on R2?

aethr: > the data in the R2 bucket is only the semi-static data that is updated semi-annually. I'm fine if these data are updated on an ad hoc basis rather than being checked monthly and automatically used assuming @prayner is also. The step of "productionising" those semi-static data might be as simple as documentation (public or private).

@lewisjared that is my understanding as well. As I understand it the data in the R2 bucket are all inputs to the prior which aren't know to change on a predictable cadence. @prayner will have to help us understand where these are sourced from and how we can understand when they might need to be updated.

> The [GFAS data](https://github.com/openmethane/openmethane-prior/blob/main/omGFASEmis.py) is downloaded during the processing of its layer and will be updated at least monthly.

I wasn't aware of that, and perhaps that's an argument that this should be done on a schedule, ignoring when the "semi-static" inputs are updated.

> When you say "prior" are you referring to this repository as a whole or just the data on R2?

I understood "the prior" to be a data set we compute in advance, based on semi-static data, which forms one of the inputs to the Open Methane model. The data in R2 is the input (along with GFAS), and the code in this repo transforms and combines the input to give us data in "our format" (ie, the grid).

Not sure what the plan is for the outputs, but if this only needs to get run once each time data changes (ie, new GFAS) it does seem like a good idea to store/cache it somewhere the model can access it.

Sorry, my comments probably aren't that helpful! Will wait for @prayner to illuminate things. :)

lewisjared: Comments are always helpful. We are all trying to get on the same page and use the same language.

I think your description of the prior is correct, other than adding that this is our initial guess at the methane emissions across Australia.

I think that this has to be run at least monthly to pull in the new GFAS (and maybe wetlands) data. In our meeting notes, we said that the "above steps are run daily" which included the prior calculation so it might even need to be run at a daily cadance. The current implementation allows for individual layers to be rerun as needed so this isn't a major can be heavily cached.

aethr: > the "above steps are run daily" which included the prior calculation so it might even need to be run at a daily cadance.

If that is true then I think "the prior" is probably the process (ie "run the prior") rather than the data set (ie "access the prior").

PR #3, opened 2024-04-08
Run ruff for an example file

lewisjared:
## Description

Runs ruff over `omUtils.py` to automate consistent formatting and linting.

[Ruff docs](https://docs.astral.sh/ruff/)
[Ruff GitHub](https://github.com/astral-sh/ruff)

The command to run the linting and reformatting is as follows:

```
ruff check --fix               
ruff format
```

I also pulled out some dead code which was no longer referenced inside the project. I'm assuming that this repository isn't installed anywhere else?

We typically use a MakeFile in our repos to run these types of commands in combination with a [pre-commit hook](https://pre-commit.com/) that runs on every commit to enforce the code style. The pre-commit hook means that you don't have to really think about the styling, as ruff will reformat everything automatically.

@prayner I notice that you use camelCase variable naming instead of the more common snake_case convention. Do you have a strong preference for that?

## Checklist

- ~~[ ] Tests added~~
- ~~[ ] Documentation added (where applicable)~~
 

## Notes

Ruff pulls together the linting rules from a number of different projects (flake8, pydocstyle,...) into a single package and also provides black-compatible formatting. The developer-experience is so much better than it was a few years ago.


lewisjared:
Can ruff include a space after all paranetheses?

lewisjared:
CamelCase will result in better UX for @prayner

aethr:
This branch was deleted successfully. :thumbsup:

PR #6, opened 2024-05-06
Add tests

crdanielbusch:
I'm testing the github workflow with this pull request. So it can be ignored for now.


.github/actions/setup/action.yaml
```diff
@@ -0,0 +1,49 @@
...
+inputs:
+  python-version:
+    description: "Python version to use"
+    required: true
+  cache-dependency-path:
+    description: "Requirement files to install. Can be a glob pattern."
+    default: '**/requirements*.txt'
+  venv-dir:
+    default: '.venv'
+  install-cmd:
```
lewisjared, line 13: I don't think that this input is used anywhere so it can be removed.

.github/actions/setup/action.yaml
```diff
@@ -0,0 +1,49 @@
...
+      if: steps.cache-venv.outputs.cache-hit != 'true'
+      shell: bash
+
+    - run: |
+        source ${{ inputs.venv-dir }}/bin/activate
+        python -m pip install -r requirements.txt
+        python -m pip install -r requirements-dev.txt
+        echo "VIRTUAL_ENV=${VIRTUAL_ENV}" >> $GITHUB_ENV
+        echo "${VIRTUAL_ENV}/bin" >> $GITHUB_PATH
+        cp .env.example .env
```
lewisjared, line 48: This line should be moved into its own step. Currently, it will be cached until the requirements files change which might lead to odd behaviour

lewisjared (commented):
Urgh I see what you mean.

In this case, it is fine to wrap download part of the `omDownloadInputs.py` in a function and then call that function from within a `if __name__ == '__main__':` block. That will allow you to import `omDownloadInputs` without triggering the download and follows a pattern that is similar to other scripts in the repo.

That should allow you to not need the `temporary_file_for_tests.py`

PR #8, opened 2024-05-13
Run om prior

crdanielbusch:
## Description
- [x] Get the calculation of emission for all layers running again
- [x] Add a test that checks the full process is running
- [x] Speed up the process to calculate emissions for Agriculture, LULUCF and waste layer.

## Checklist

- [x] Tests added
 

## Notes

@prayner When we check the data with `omPriorVerify.py`, we get the following output. Does that still work?

```
PASSED - Livestock CH4 within bounds of total agriculture CH4: 0.16963856
FAILED - Discrepency of -51.030272990142386% in agriculture emissions
FAILED - Discrepency of -65.28372899999995% in electricity emissions
FAILED - Discrepency of -89.48823911165206% in fugitive emissions
FAILED - Discrepency of -73.59460500000147% in industrial emissions
FAILED - Discrepency of -77.15502137254902% in lulucf emissions
FAILED - Discrepency of -73.59460972972967% in stationary emissions
FAILED - Discrepency of -73.59461375000018% in transport emissions
FAILED - Discrepency of -79.06976740181267% in waste emissions
```


tests/conftest.py
```diff
@@ -0,0 +1,6 @@
+# Arrange
```
lewisjared, line 1: Is this a dangling sentence?

tests/conftest.py
```diff
@@ -0,0 +1,6 @@
...
+import pytest
+from pathlib import Path
+@pytest.fixture
```
lewisjared, line 4:
```suggestion


@pytest.fixture
```

Once we have ruff'ed the codebase we should add some pre-commit hooks to ensure the formatting is always applied

lewisjared (approved):

cmaq_example/GRIDCRO2D_1
aethr, line 1: I had a look at these in Panoply, and although I barely understand NetCDF files, one thing I noticed is that both `cmaq_example/GRIDCRO2D_1` and `cmaq_example/GRIDDOT2D_1` have the same values for:

```
  :XCELL = 25000.0; // double
  :YCELL = 25000.0; // double
```

I think this corresponds to the 25km grid, which means these files are a bit outdated.

@prayner has commented that both of these files should be outputs from `setup-wrf`. Although I didn't see these from running `setup_for_wrf.py` in the docker container, I did notice it generates another NetCDF file in that repo: `data/runs/aust-test/2024010100/geo_em.d01.nc` (although weirdly this ends up as a broken symlink to a file that actually exists in `templates/aust-test/geo_em.d01.nc`. In this file, we see:

```
  :DX = 10000.0f; // float
  :DY = 10000.0f; // float
```

I think this corresponds to the 10km grid, although I'm unsure why the attribute names have changed.

Lastly, I noticed there is a script in `setup-wrf` called `make_grdfile.sh`: https://github.com/openmethane/setup-wrf/blob/main/make_grdfile.sh. However, this seems tightly coupled to the location of some files on a server somewhere (perhaps nci?).

@prayner is there any chance of getting access to those files to be able to re-create the 10km grid files?

lewisjared: We just reverted an old commit to get the system running. It's amazing that this ran if they are for different domains... @crdanielbusch We need to add a check that the CRO and domain files are consistent. 

`make_grdfile.sh` requires a working CMAQ container so we aren't up to that part of the pipeline yet

prayner:
I tried to send links to 10km versions of these files yesterday,
clearly that didn't work. You can find these at
https://www.dropbox.com/scl/fi/h8quykmt9u16ggmmv8zjr/GRIDDOT2D_2?rlkey=sr5ka3rjp8y3vkew8r2lln6zb&dl=0
https://www.dropbox.com/scl/fi/hyd7tdk3ows3asytzxl4y/GRIDCRO2D_2?rlkey=jn2kj9lxtkycsfeat1atx1m4w&dl=0
The geo_em file is produced by the wps process before running WRF, the
GRID* files are produced in the MCIP process run as part of the
setupCMAQinputs script and is generated from the wrfout files. They're
pretty similar except that the GRID files have some edge gridcells
removed since the meteorology here is less reliable. I'd be very
surprised if the code will run with different resolutions.

cmaq_example/GRIDCRO2D_1
crdanielbusch: > @crdanielbusch We need to add a check that the CRO and domain files are consistent.

I have added a test that checks if the grid size (`DX` and `DY`) is 10 km in `geo_em.d01.nc`, `GRIDCRO2D_1` and `GRIDDOT2D_1`. 

What other files should be compared? `inputs/om-domain-info.nc` and `GRIDCRO2D_1`?

crdanielbusch:
Thanks for sending the files @prayner. They are now in the repository. 

The tests in `omPriorVerify.py` now pass.

lewisjared:
Great! We should refactor the verify script into something that can be tested in a future MR

prayner:
I'm wary of including the GRID* files in the repo since they're not
really input files nor are they part of the prior process itself but
are used to create om-domain-info.nc. Can we perhaps include
om-domain-info.nc itself?

prayner:
In my normal workflow om-domain-info is produced from the GRID* files
then the GRID* files are not used again. If omCreatDomainInfo runs
successfully we should (a) have consistent grids and (b) not need the
GRID* files again. If the prior is only being developed for Open
Methane then none of this matters much. 
-- 
Peter Rayner (he/him), Honorary Professorial Fellow, University of Melbourne
mobile +61 402 752 379
mail-to: ***@***.***

PR #14, opened 2024-05-21
Refactor `omPriorVerify.py` into tests

crdanielbusch:
## Description

- [x] Remove `omPriorVerify.py` call from `omPrior.py`
- [x] Refactor tests from `omPriorVerify.py` in test suite - see tests 005 and 006
- [x] Combine all tests in one module
- [x] Add a fixtures that downloads input files
- [x] Add a fixture that runs the whole process and references the input files fixture
- [x] Fix CI
- [x] Add a time out for the tests (currently at 10 min)

## Checklist

- [x] Tests added


tests/test_om_prior.py
```diff
@@ -1,32 +1,175 @@
...
+    agVal = round(sector_data["agriculture"] * 1e9)
+    agDX = agVal - lsVal
+
+    assert agDX > 0, f"Livestock CH4 exceeds bounds of total agriculture CH4: {agDX / 1e9}"
+
+
+# TODO Update this test when file structure is clear.
+# This test ensures that the grid size for all input files is 10 km.
+def test_006_grid_size_for_geo_files(root_dir, monkeypatch) :
+    expected_cell_size = 10000
```
crdanielbusch, line 125: I am not sure what exactly to test here. Right now the test ensure that the grid size for `geo_em.d01.nc`, `GRIDCRO2D_1`, `GRIDDOT2D_1` is 10 km.

tests/test_om_prior.py
```diff
@@ -1,32 +1,175 @@
...
+
+    yield xr.load_dataset(filepath_out_domain)
+
+    os.remove(filepath_out_domain)
+
+    os.remove(filepath_in_domain)
+
+
+# Fixture to download and later remove all input files
+@pytest.fixture(scope="session")
```
lewisjared, line 43: This fixture can probably move into `conftest.py` as it could be used in other places. Also since it is session scoped even if you use it in multiple test modules it will still only run once

tests/test_om_prior.py
lewisjared: My initial thought was lat/lons, but then I realised that the GRIDCRO/GRIDDOT are a subset of the WRF domain.

That could be a test for after `omCreateDomainInfo.py` is run, but not applicable here

scripts/omPrior.py
```diff
@@ -57,5 +57,3 @@
 omWetlandEmis.processEmissions(args.startDate, args.endDate)
 
 omOutputs.sumLayers()
-
-omPriorVerify.verifyEmis()
```
lewisjared, line 61: I think that calling this is as part of the prior creation is still correct. The tests are in independent check that the emissions sum up correctly for the example domain.

tests/test_om_prior.py
```diff
@@ -1,32 +1,175 @@
...
+from openmethane_prior.omUtils import getenv, secsPerYear
+
+root_path = Path(__file__).parent.parent
+sys.path.insert(1, os.path.join(root_path, "scripts"))
+from omDownloadInputs import download_input_files, downloads, remote, sectoralEmissionsPath
+
+
+@pytest.fixture(scope="session")
+def output_domain_file(root_dir,
+                       input_files
```
lewisjared, line 23: This doesn't look ruffed

tests/test_om_prior.py
```diff
@@ -1,32 +1,175 @@
...
+                       ) :
+    subprocess.run(["python", os.path.join(root_dir, "scripts/omCreateDomainInfo.py")], check=True)
+
+    subprocess.run(["python", os.path.join(root_dir, "scripts/omPrior.py"), "2022-07-01", "2022-07-02"], check=True)
+
+    # Generated by scripts/omCreateDomainInfo.py
+    filepath_in_domain = os.path.join(root_dir, "inputs/om-domain-info.nc")
+
+    # Generated by scripts/omPrior.py
+    filepath_out_domain = os.path.join(root_dir, "outputs/out-om-domain-info.nc")
```
lewisjared, line 33: Can we add a new test that verifies that this output file has the same lat/lons that are in the CRO/DOT files. `filepath_in_domain` should have the same dimensions as well.

lewisjared (approved):
Looks good. There are a few suggestions for other things to test. Whether this comes in the PR or later is up to you.

.github/actions/setup/action.yaml
```diff
@@ -49,3 +49,5 @@ runs:
         echo "${VIRTUAL_ENV}/bin" >> $GITHUB_PATH
 
       shell: bash
+
+
```
lewisjared, line 53:
```suggestion
```

prayner:
I'm worried about hard coding a 10km test since subdomains will be
finer resolution
-- 
Peter Rayner (he/him)
My work days are Monday-Thursday, responses at other times might be slow.
The Superpower Institute <https://www.superpowerinstitute.com.au/>
Level 3 105 Victoria St. Fitzroy 3065
Transparent, actionable and auditable data on Australia's methane emissions <https://openmethane.org>
mobile +61 402 752 379
mail-to: ***@***.***

lewisjared:
> I'm worried about hard coding a 10km test since subdomains will be
finer resolution

@prayner In this case we are testing that the example domain that is checked into the repo is consistent. These tests aren't run for every new domain (maybe they should eventually)

tests/test_om_prior.py
crdanielbusch: Is it possible that `GRIDCRO2D_1` and `GRIDDOT2D_1` have different dimensions?  

In this test I compare `COL` and `ROW` of `GRIDCRO2D_1` and `GRIDDOT2D_1` with `inputs/om-domain-info.nc`. 

The number of columns and rows  in `GRIDCRO2D_1` is one below the number of columns and rows in `GRIDDOT2D_1`:

`GRIDCRO2D_1` contains the attributes `ROW: 430, COL: 454`

and 

`GRIDDOT2D_1` contains the attributes `ROW: 431, COL: 455`.

In `om-domain-info.nc` there are the attributes `ROW: 430` and `ROW_D: 431`.
So I can compare `ROW` with  `GRIDCRO2D_1` and `ROW_D` with `GRIDDOT2D_1`?


The test that I am adding looks like this. It's passing, but I am not sure if the test makes sense.

```python
def test_007_compare_in_domain_with_cro_dot_files(input_domain_xr, cro_xr, dot_xr):

    assert dot_xr.NCOLS == input_domain_xr.COL_D.size
    assert dot_xr.NROWS == input_domain_xr.ROW_D.size

    assert cro_xr.NCOLS == input_domain_xr.COL.size
    assert cro_xr.NROWS == input_domain_xr.ROW.size
```

lewisjared: WRF and CMAQ use a staggered grid (https://amps-backup.ucar.edu/information/configuration/wrf_grid_structure.html) where depending on the quantity that you are measuring the grid points are slightly different. Things like temperature/pressure are defined at the center of a grid cell, where as other quantities such as wind are defined at the edges of a grid cell. If you count the number of edges in a grid along a x or y it is one larger than the number of cells in x or y.

The CRO/DOT files are generated by MCIP (?) from the WRF domain and define the domain used by CMAQ.
The CRO files define the latitude centers where as the DOT files define the corners hence why `ROW_D = ROW + 1`. I don't know why they are named as they are 🤷 

The CMAQ domain is a little smaller than the WRF domain because the 5 grid cells around the edge of the domain are excluded. I can talk you through why that is another day.

lewisjared: According to this [website](https://www.cmascenter.org/ioapi/documentation/all_versions/html/GRIDS.html), CRO is short for cross and dot. 

An image of the CMAQ grid: https://www.cmascenter.org/ioapi/documentation/all_versions/html/GRIDS.jpg
and the interaction between the grid and "thickened boundary" https://www.cmascenter.org/ioapi/documentation/all_versions/html/THKBDY.jpg. The outside is the WRF domain, but the inner cells labelled "main grid" is the CMAQ grid

crdanielbusch: Ok thanks. I think I have understood it to some degree. 

Is the difference between CMAQ domain and WRF domain (the thickened boundary) unrelated to mass grid / staggered grid problem? And it wouldn't be relevant to the tests here anyway, would it?

I've added the tests 007 and 008.

lewisjared: Correct. It is unrelated. If you compared the grid sizes in the raw input domain `geo_em.d01.nc` you would see that they don't align with the CMAQ grid

PR #22, opened 2024-05-30
Test grid

lewisjared:
## Description

Use the test grid generated from setup-wrf (10x10 grid with 10km spacing) instead of the full 10km Aus domain

## Checklist

- [ ] Tests added


lewisjared:
Hmm this might not be as good an idea as I initially thought. Given the small size of the domain several layers are zero

lewisjared:
@crdanielbusch If we get fed up with the slow CI we could think about making this work

PR #32, opened 2024-07-01
Export all layers of the prior in omGeoJSON.py

aethr:
## Description

This script is still just used as a debugging script, but exporting all layers gives us more data to debug with.

On one hand, we can visualise each layer distinctly in the front-end, allowing each layer to be checked visually. As a secondary benefit, each layer provides a distinct but realistic example of methane emissions that can be used as mock data for front-end development.

## Checklist

- [] ~~Tests added~~
- [] ~~Documentation added (where applicable)~~

Note: I have not added tests, as this file is still only useful for debugging and I do not believe it forms part of the production workflow.

## Notes

I attempted to update this script to work more like omDomainJSON.py which uses xarray to select the specific NetCDF layers we're interested in, however that approach wasn't easy because the variables in the prior output file don't all share the same dimensions.

### DIFF
--- a/scripts/omGeoJSON.py
+++ b/scripts/omGeoJSON.py
@@ -91,7 +91,7 @@ def processGeoJSON():
             # raw values
             properties[layer_name] = float(ds_slice[layer_name][y][x])
             # relative to max
-            properties[f"{layer_name}_R"] =  float(ds_slice[layer_name][y][x] / max_values[layer_name]),
+            properties[f"{layer_name}_R"] = float(ds_slice[layer_name][y][x] / max_values[layer_name])
 
         features.append(
             Feature(
--- a/scripts/omGeoJSON.py
+++ b/scripts/omGeoJSON.py
@@ -25,6 +25,21 @@ import numpy as np
 from geojson import Feature, FeatureCollection, Polygon, dumps
 from openmethane_prior.omOutputs import domainOutputPath, geoJSONOutputPath
 
+prior_layers = [
+    "OCH4_AGRICULTURE",
+    "OCH4_LULUCF",
+    "OCH4_WASTE",
+    "OCH4_LIVESTOCK",
+    "OCH4_INDUSTRIAL",
+    "OCH4_STATIONARY",
+    "OCH4_TRANSPORT",
+    "OCH4_ELECTRICITY",
+    "OCH4_FUGITIVE",
+    "OCH4_TERMITE",
+    "OCH4_FIRE",
+    "OCH4_WETLANDS",
+    "OCH4_TOTAL",
+]
 
 class NumpyEncoder(json.JSONEncoder):
     """Numpy encoder for JSON serialization"""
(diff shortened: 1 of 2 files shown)


scripts/omGeoJSON.py
```diff
@@ -42,48 +42,111 @@ def processGeoJSON():
...
+                    "OCH4_INDUSTRIAL": float(ds_slice["OCH4_INDUSTRIAL"][y][x]),
+                    "OCH4_STATIONARY": float(ds_slice["OCH4_STATIONARY"][y][x]),
+                    "OCH4_TRANSPORT": float(ds_slice["OCH4_TRANSPORT"][y][x]),
+                    "OCH4_ELECTRICITY": float(ds_slice["OCH4_ELECTRICITY"][y][x]),
+                    "OCH4_FUGITIVE": float(ds_slice["OCH4_FUGITIVE"][y][x]),
+                    "OCH4_TERMITE": float(ds_slice["OCH4_TERMITE"][y][x]),
+                    "OCH4_FIRE": float(ds_slice["OCH4_FIRE"][y][x]),
+                    "OCH4_WETLANDS": float(ds_slice["OCH4_WETLANDS"][y][x]),
+                    "OCH4_TOTAL": float(ds_slice["OCH4_TOTAL"][y][x]),
+                    "m": float(ds_slice["OCH4_TOTAL"][y][x]),
```
aethr, line 123: Left for backwards compatibility with previous verison.

scripts/omGeoJSON.py
```diff
@@ -42,48 +42,111 @@ def processGeoJSON():
...
+                    "OCH4_INDUSTRIAL_R": float(ds_slice["OCH4_INDUSTRIAL"][y][x] / max_values["OCH4_INDUSTRIAL"]),
+                    "OCH4_STATIONARY_R": float(ds_slice["OCH4_STATIONARY"][y][x] / max_values["OCH4_STATIONARY"]),
+                    "OCH4_TRANSPORT_R": float(ds_slice["OCH4_TRANSPORT"][y][x] / max_values["OCH4_TRANSPORT"]),
+                    "OCH4_ELECTRICITY_R": float(ds_slice["OCH4_ELECTRICITY"][y][x] / max_values["OCH4_ELECTRICITY"]),
+                    "OCH4_FUGITIVE_R": float(ds_slice["OCH4_FUGITIVE"][y][x] / max_values["OCH4_FUGITIVE"]),
+                    "OCH4_TERMITE_R": float(ds_slice["OCH4_TERMITE"][y][x] / max_values["OCH4_TERMITE"]),
+                    "OCH4_FIRE_R": float(ds_slice["OCH4_FIRE"][y][x] / max_values["OCH4_FIRE"]),
+                    "OCH4_WETLANDS_R": float(ds_slice["OCH4_WETLANDS"][y][x] / max_values["OCH4_WETLANDS"]),
+                    "OCH4_TOTAL_R": float(ds_slice["OCH4_TOTAL"][y][x] / max_values["OCH4_TOTAL"]),
+                    "rm": float(ds_slice["OCH4_TOTAL"][y][x] / max_values["OCH4_TOTAL"]),
```
aethr, line 139: Left for backwards compatibility with previous verison.

scripts/omGeoJSON.py
```diff
@@ -42,48 +57,65 @@ def processGeoJSON():
...
     ds = nc.Dataset(domainOutputPath)
-    landmask = ds["LANDMASK"][:]
-    lats = ds["LATD"][:][0]
-    longs = ds["LOND"][:][0]
-    ch4 = ds["OCH4_TOTAL"][:][0][0]
-    maxEmission = np.amax(ch4)
+
+    # There is a better way to do this but this will work for now
+    # Using xarray wasn't straightforward because the layers don't use
+    # the same attribute names, ie mixing x/y and lat/long.
```
lewisjared, line 63: That needs to be fixed. The naming is incorrect, but the dimensions are the right size which makes it harder to handle than it should

scripts/omGeoJSON.py
```diff
@@ -42,48 +57,65 @@ def processGeoJSON():
...
+        "LATD": ds["LATD"][:][0][0],
+        "LOND": ds["LOND"][:][0][0],
+    }
+
+    max_values = {}
+    for layer_name in prior_layers:
+        # extract the meaningful dimensions from the NetCDF variables
+        ds_slice[layer_name] = ds[layer_name][:][0][0]
+        # find the max emission value in a single cell for each layer
+        max_values[layer_name] = np.amax(ds_slice[layer_name])
```
lewisjared, line 75:
```suggestion
        max_values[layer_name] = np.max(ds_slice[layer_name])
```
Max and amax are the same. I had to look up what amax was haha

lewisjared (approved):

scripts/omGeoJSON.py
aethr: Haha all good, it was like that when I found it!

Updated in 2c7edba.

scripts/omGeoJSON.py
aethr: Agree, but not within the scope of this change. The code here should continue to work after layer names are changed due to its simplicity, so I'd like to leave it as-is for now.

lewisjared: 👍

PR #33, opened 2024-07-08
Clean the global state

lewisjared:
## Description
Currently a lot of global state is used (inputs, outputs etc). This makes it very hard to test the application and generally requires running scripts via `subprocess` which is hard to debug.

This removes the global state in preference for a `PriorConfig` object to carry around the options. This config object is still created from environment variables.

This also enables the GeoJSON test which was previously disabled because of dependency issues.

## Checklist

- [x] Tests added
 

## Notes

This also moves the input and output directories into the `data/` directory


lewisjared:
@crdanielbusch Ping

crdanielbusch (approved):
Nice, looks much cleaner now.

I've tested `make run` and `make download` on my machine, it all works fine.

Issue #16, opened 2024-05-28
Unit of methane emissions


NasimehShahrokhi:
## The motivation

The unit of emissions is not recorded in the generated output emission files.

lewisjared:
Units were added to the output dataset in #36. Units are always `kg/m^2/s`

PR #36, opened 2024-07-09
adding unit and name attributes to layers

prayner:
addresses #16

-------

## Checklist

- [ ] Tests added
 

## Notes


lewisjared (commented):
I think we need to add the required attributes to the total as well as the layers in the `sum_layers` function.

We should also add a test to the `test_om_prior.py` which verifies that the units are in the output. This helps to ensure that they don't get lost if we do further changes. This one is a bit annoying because it requires running the whole workflow to test this so it can be a bit slow.

A simple test like this could be sufficient for now:

```
def test_required_attributes(output_domain_xr):
    assert output_domain_xr.variables["OCH4_TOTAL"].attrs == {
        "units": "kg/m^2/s",
        "long_name": "OCH4_TOTAL",
    }
    assert output_domain_xr.variables["OCH4_WETLANDS"].attrs == {
        "units": "kg/m^2/s",
        "long_name": "OCH4_WETLANDS",
    }
```

This currently fails until the fix to the total has been implemented.

We should have a session on how to run [pytest](https://docs.pytest.org/en/8.2.x/) and the test suite.

prayner:
added the attributes, I'll leave the test to someone who understands pytest

lewisjared:
@prayner Are you able to see any line-specific comments or are they not available in emacs?

prayner:
Don't think I can see them in emacs ... let's see if the gh cli
reveals them

prayner:
I'm not seeing in-line comments, are there some here?

lewisjared (approved):
Ok good to know.

For constants, "SCREAMING_SNAKE_CASE" is the preference. If we don't change this as part of this PR it will likely later be cleaned up.

I had another comment to use `setncatts` to set all the values at the same time instead of looping.

I'll fix them up with the tests

lewisjared:
@prayner Can I add the tests to this branch? If so please assign this PR to me

PR #37, opened 2024-07-11
Use published domain

lewisjared:
## Description

Use the published versions of the openmethane domains for testing.

Adds `DOMAIN_NAME` and `DOMAIN_VERSION` environment variables to specify the domain name and version to use.
The appropriate domain is then downloaded with the other layer data using `scripts/omDownloadInputs.py`. This file is then copied to the output directory when running `omPrior`

Adds `OUTPUT_DOMAIN` environment variable to specify the name of the output domain file.

## Checklist

- [x] Tests added


crdanielbusch (approved):

crdanielbusch:
Is it the possible that the make commands `clean`, `clean-all` and `run` still need to be updated? 

```
clean:  ## remove generated temporary files
	find intermediates outputs -type -delete
```

lewisjared:
Nice catch. Can we document the data folders in the README instead?

crdanielbusch:
> Nice catch. Can we document the data folders in the README instead?

Yep, I can add it

PR #65, opened 2025-01-20
Resolving some wrong lines in README as well as adding some more info…

Ida-droid:
## Description

making documentation clearer
## Checklist

- [ ] Tests added

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,9 +23,25 @@ Note: the ADS API is different from the CDS (Climate Data Store) API
 even though they are both parts of the Copernicus program
 and share the same credentials file.
 
+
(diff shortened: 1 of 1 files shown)


aethr (approved):
Looks great, thank you for the contribution @Ida-droid !

aethr:
Tasks failing in CI is likely to be due to secrets not being accessible to the workflow, since it wasn't triggered by a team member.

In this case none of the checks matter and we don't need containers to be built since the change was only to the README. It will take some investigation in the future if we get pull requests from non-team members with code changes, where we'll need CI/CD to run.

PR #70, opened 2025-06-23
Adopt CF Conventions in prior output

aethr:
## Description

Similar to https://github.com/openmethane/openmethane/pull/117, this PR updates the output format for the prior to adopt some [CF Conventions](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.12/cf-conventions.html).

We'll also use this opportunity to adopt clearer naming conventions and file organisation to make the output and the project more approachable.

### Output structure and naming

To summarise the discussion in the comments below, the output format we're aiming for would be:

- coords
  -`x` and `y` spatial coordinates
  - `time` - one entry for each day, with `time_bounds` recording the start and end of each period
  - `vertical` a single-coordinate dimension which can be extended later
- data
  - `land_mask` (variable) - normalised copy of `LANDMASK`
  - `ch4_total` (variable) - normalised copy of `OCH4_TOTAL` - sum of all sectors
  - `ch4_sectors` (group)
    - `agriculture`
    - `electricity`
    - `fugitive`
    - `industrial`
    - `lulucf`
    - `stationary`
    - `transport`
    - `waste`
    - `livestock`
    - `fire`
    - `wetlands`
    - `termite`

### Sectors `standard_name`

We will adopt CF Conventions `standard_name` for each emission variable:
- `ch4_total`: `"surface_upward_mass_flux_of_methane"`
- `ch4_sectors.agriculture`: `"surface_upward_mass_flux_of_methane_due_to_emission_from_agricultural_production"`
- `ch4_sectors.electricity`: `"surface_upward_mass_flux_of_methane_due_to_emission_from_energy_production_and_distribution"`
- `ch4_sectors.fugitive`: `"surface_upward_mass_flux_of_methane_due_to_emission_from_extraction_production_and_transport_of_fuel"`
- `ch4_sectors.industrial`: `"surface_upward_mass_flux_of_methane_due_to_emission_from_industrial_processes_and_combustion"`
- `ch4_sectors.lulucf`: `"surface_upward_mass_flux_of_methane_due_to_emission_from_anthropogenic_land_use_change"`
- `ch4_sectors.stationary`: `"surface_upward_mass_flux_of_methane_due_to_emission_from_industrial_energy_production"`
- `ch4_sectors.transport`: `"surface_upward_mass_flux_of_methane_due_to_emission_from_land_transport"`
- `ch4_sectors.waste`: `"surface_upward_mass_flux_of_methane_due_to_emission_from_waste_treatment_and_disposal"`
- `ch4_sectors.livestock`: `"surface_upward_mass_flux_of_methane_due_to_emissions_from_domesticated_livestock"`
- `ch4_sectors.fire`: `"surface_upward_mass_flux_of_methane_due_to_emission_from_fires"`
- `ch4_sectors.wetlands`: `"surface_upward_mass_flux_of_methane_due_to_emission_from_wetland_biological_processes"`
- `ch4_sectors.termite`: `"surface_upward_mass_flux_of_methane_due_to_emission_from_termites"`

## Tasks

- [x] create output dataset which follows basic CF conventions attributes
- [x] rename `ROW` and `COL` coordinates to `y` and `x`
- [x] rename `LAY` coordinate to `vertical`
- [x] migrate from `TSTEP` or `date` coordinates to `time`
- [x] expand all layer outputs to include every time step for consistent `time` coordinate
- [x] duplicate `LANDMASK` as normalised `land_mask` following CF conventions
- [x] duplicate `OCH4_TOTAL` variable as `ch4_total`
- [x] prefix sectoral layer variables with `ch4_sector_`
- [x] apply consistent attributes like "standard_name" and "units" to every output variable (see above)

## Checklist

- [x] Tests added
- [x] Create follow up tasks for:
  - [x] migrate downstream uses of `OCH4_TOTAL` to `ch4_total`
  - [x] remove `OCH4_TOTAL` variable
  - [x] migrate downstream uses of `LANDMASK` to `land_mask`
  - [x] remove `LANDMASK` variable

## Notes

In 9e1d38abda606993c777788133510baba260c6d0 I re-generated the `tests/integration/test_om_prior/test_009_output_domain_xr.csv` test fixture before making any changes to the prior output. This is because the pytest-regression module does allow differences between floats between test runs due to differences handling floats across different hardware. However, I wanted to be able to compare the results in this file before and after making changes to the output format, to ensure that my changes had no effect on the (numerical) scientific results.


aethr:
Questions for @prayner that arise from the naming exercise in the previous comment:

1. CF standard names includes some names like "due_to_emission_from_fires_excluding_anthropogenic_land_use_change" which suggests that the LULUCF category may include emission from intentional fires due to land use change. I would imagine that GFAS doesn't discriminate between wildfires and intentional fires. Does the LULUCF layer include emissions from fires, and if so could we be double counting these?

2. Would it be a problem for the "OCH4_INDUSTRIAL" and "OCH4_STATIONARY" layers to share the same standard name? If so, do you have a suggestion for names for those two layers? They seem aligned with NGGI sectoral accounting.

3. Does the OCH4_WETLANDS layer represent methane production only, or does it also account for methane consumption (ie production - consumption)?

prayner:
>Adopting CF Conventions does mean attempting to classify each data variable in our output
>with a standard_name from the CF Standard Name Table / conventions.
>
>Broadly, all of our methane emission layers will share a naming convention like:
>
>* standard_name: "{standard_name for emission}_due_to_emission_from_{standard_name for source}"
Agreed.
>
>Naming emissions
>
>The candidates for standard_name for emission seem to be:
>
>* surface_upward_mass_flux_of_methane 
>
> * CF conventions description: 
>
> * "Upward" indicates a vector component which is positive when directed upward
> (negative downward). A net upward flux is the difference between the flux from below
> (upward) and the flux from above (downward).
There's only one case I can think of where the "net" matters.
Otherwise all our fluxes are entirely upward from below. I suggest
emphasizing this by not using "net" except for the one case where it matters.
> * While "upward mass flux" does seem accurate in describing what we're estimating, the
> "surface" part is less clear. Would it be fair to say that methane emissions we're
> estimating occur at or near to the "surface". I.e., methane from fires, grazing
> animals, wetlands, industrial activity "enters the atmosphere" at the "surface"?
In the current version this is true but it's not obligatory i.e CMAQ
can define emissions on any layer. As we move to higher resolution this
may well happen. I would still keep the "surface" in the name until it
does: CMAQ ingests emissions by inserting them throughout the layer
they occur in, even though they enter at one edge of the layer,
usually the bottom. For CMAQ it makes no difference but for other
models it does so I think we should keep the exact emission point
clear for . 
>
>* tendency_of_atmosphere_mass_content_of_methane 
Provided it's associated with a particular process so ... due to
emissions ... it's ok but that seems unnecessarily compicated when the
emissions are in the same unit. Terms like this are often used for
chemical destruction of methane ... not our problem here.

>Naming layer sources
>
>Candidates for layer sources might be:
I can't remember why someone added the "O" at the start.
>* OCH4_AGRICULTURE - "due_to_emission_from_agricultural_production"
>* OCH4_LULUCF - "due_to_emission_from_anthropogenic_land_use_change"
>* OCH4_WASTE - "due_to_emission_from_waste_treatment_and_disposal"
>* OCH4_LIVESTOCK 
>
> * "due_to_emission_from_grazing"
Yes, we want that, curious that agriculture is listed at the same
level or is it a higher level category? Within openmethane prior we
currently have an agriculture category and subtract the livestock. 
> * "due_to_emission_from_herbivorous_mammals"
The risk there is it might include non-farmed animals. Not a problem
in Australia (kangaroos don't make methane) but elsewhere yes. The
only problem with the "grazing" term is agricultural people might not
think animals in feedlots are grazing. I would have just said
"due_to_emissions_from_domesticated_livestock". As a note I think we
only have ruminants in the Australian database we use so pork and
poultry are included with the rest of agriculture. 
>
>* OCH4_INDUSTRIAL - "due_to_emission_from_industrial_processes_and_combustion"
>* OCH4_STATIONARY - "due_to_emission_from_industrial_processes_and_combustion" 
>
> * is it ok to be the same as INDUSTRIAL?
I think what's happening here is one is energy produced and consumed
*within* an industrial plant and the othher is any other emissions
from the industrial process. I think the industrial energy is
sometimes called industrial_autoproduction.

https://www.ipcc-nggip.iges.or.jp/public/2006gl/pdf/1_Volume1/V1_8_Ch8_Reporting_Guidance.pdf.
I would suggest:
OCH4_STATIONARY - "due_to_emission_from_industrial_energy_production" 
>
>* OCH4_TRANSPORT - "due_to_emission_from_land_transport"
>* OCH4_ELECTRICITY - "due_to_emission_from_energy_production_and_distribution"
>* OCH4_FUGITIVE - ??? 
1 B Fugitive Emissions from Fuels Includes all intentional and unintentional emissions from the
extraction, processing, storage and transport of fuel to the
point of final use.
So I would suggest:
OCH4_FUGITIVE - "due_to_emission_from_extraction_production_and_transport_of_fuel"
>
>* OCH4_TERMITE - "due_to_emission_from_termites"
>* OCH4_FIRE - "due_to_emission_from_fires"
>* OCH4_WETLANDS 
>
> * "due_to_emission_from_wetland_biological_processes" - difference between methane
> production and consumption
> * "due_to_emission_from_wetland_biological_production" - only includes methane production
We want the first one because our emissions there do involve removals
when the land is a sink.

aethr:
Thanks @prayner I will action all of those names.

> I can't remember why someone added the "O" at the start.

I was going to ask about this. "OCH4" suggests methanol (doesn't it?), which I find confusing.

Now is a good opportunity to change the names, both for the layers (which have no downstream consumer in Open Methane) and for the total. I'll leave "OCH4_TOTAL" intact until I can update openmethane downstream, but we can duplicate it with a "better" name in the meantime.

In openmethane's `posterior_emissions.nc` we use "CH4" and "prior_CH4", which makes sense in that context. I guess "CH4" or "CH4_total" or "total_CH4" might make sense for the output of the prior?

For the layers, I was thinking of trying to adopt DataTree / NetCDFv4 "group" to segregate the total and the parts, like:

- CH4 (variable)
- CH4_sources (group)
  - agriculture (variable)
  - lulucf (variable)
  - waste (variable)
  - etc.

What's the clearest name for the group (ie "CH4_sources")? I'm not sure if "layer" is the best term, since that often means vertical layer in a 3D grid. Also they aren't exactly layers in the sense that they don't have a spatial relationship other than occurring in the same cell, they are parts or components that compose to make up the total.

prayner:
>1 CF standard names includes some names like
> "due_to_emission_from_fires_excluding_anthropogenic_land_use_change" which suggests that
> the LULUCF category may include emission from intentional fires due to land use change. I
> would imagine that GFAS doesn't discriminate between wildfires and intentional fires.
> Does the LULUCF layer include emissions from fires, and if so could we be double counting
> these?
I doubt it but that one is worth checking. CH4 from LULUCF is pretty
small I think.
>2 Would it be a problem for the "OCH4_INDUSTRIAL" and "OCH4_STATIONARY" layers to share the
> same standard name? If so, do you have a suggestion for names for those two layers? They
> seem aligned with NGGI sectoral accounting.
See previous response
>3 Does the OCH4_WETLANDS layer represent methane production only, or does it also account
> for methane consumption (ie production - consumption)?
and again

prayner:
>1 Many of the layers currently have 4 dimensions, like: [TSTEP, LAY, ROW, COL]. Within this
> structure it seems that the "LAY" dimension always contains only a single element. Is it
> safe to remove the dimension so we get something like: [TSTEP, ROW, COL]? Will that
> impact anything downstream like CMAQ?
I'd keep it. If we do implement a plume model for fire emissions for
example these will become multi-level. CMAQ isn't a problem since
there's more processing between this file and CMAQ inputs.
>2 OCH4_FIRE and OCH4_WETLANDS end up with a time step for each day when running the prior
> over multiple days, but other layers only include a single timestep over the same period.
> If this is correct / desired, I should create a couple of different time bounds values to
> represent the 1-day periods and the multi-day periods correctly.
Yes. Both those emissions can change with time depending on the period
covered so needed a daily step. for consistency we should also have
bounds for them.

aethr:
> > For the layers, I was thinking of trying to adopt DataTree / NetCDFv4 "group" to segregate
the total and the parts, like:

> I don't see any harm in that but not much benefit either. Most of the
clutter in root came from the layers and we've shifted much of it to
the group. The only check is whether it forces duplication of
coordinate variables. If not I'd say go ahead.

I think removing clutter from the root would be useful, but more importantly it lets us delineate the set of sector layers clearly. At the moment there is a `layer_names` variable exported somewhere which manually lists all of the layers, which gets used in loops for each bit of code that wants to iterate over sectors (for example, to sum them). It would be easy to add a new layer and forget to add it to this list.

If we put all layers/sectors in a group, then we can rely on the NetCDF file to tell us what the layers are without much fuss.

> Yes. Both those emissions can change with time depending on the period
covered so needed a daily step. for consistency we should also have
bounds for them.

This creates an awkward situation in the NetCDF file, since different variables now have different time steps. I think we could solve in one of two ways:
1. we create separate coordinates/dimensions for the two different types of time step:
  - `time` - 24h time steps for each day in the period, used by fire and wetlands
  - `time_full` (name suggestions welcome) - a single coordinate dimension with time_bounds from start -> end
2. we expand the non-stepped variables to effectively copy their values n times to share the same time step coordinate as fire/wetlands. this would duplicate a lot of data, and may be misleading to consumers?

aethr:
> I'd keep it. If we do implement a plume model for fire emissions for
example these will become multi-level.

I believe the CF convention then would be to call this dimension/coordinate "vertical", is that acceptable for our purposes?

aethr:
I've attempted to condense the decisions we arrived at in the conversation above into tasks in the PR description and documented the desired file structure and `standard_name` values.

Regarding the `time` dimension, @prayner and I spoke on the phone and agreed that the simplest solution would be to expand the layers with a single value to duplicate their data across every time step present in the fire and wetlands layers. While this may create the impression that we have discrete data for each time step in all layers, Peter thinks it's not that uncommon to do this in the field, and it makes computation across multiple layers easier as they all share the same structure.

aethr:
> move sectoral layer variables into ch4_sectors DataTree / NetCDF "group"

My attempt moving sectoral data into groups was unsuccessful. There are some critical differences between xarray Dataset and xarray DataTree (which is necessary for writing/reading a file with groups) for which the documentation is not great. I did manage to get the sectors being written to variables within a group, but Panoply no longer recognised them as georeferenced variables, so would not plot them on a map.

I'm not sure whether this is a limitation with panoply, or whether my methodology was flawed, but I spent a lot of time trying to correct the issue unsuccessfully.

I spoke to Peter and we decided to abandon the group idea for now, and simply prefix the sectoral layer variables with `ch4_sector`.

aethr:
@prayner I found a tool called [cfchecks](https://github.com/cedadev/cf-checker) and ran it over the output from this PR, and it made a few suggestions to fix. I've pushed a few new commits this morning addressing those.

The most substantial is that the "x" and "y" coordinates are now based on the raw projection coordinates in meters. We discussed previously leaving the coordinates as grid cell coords (0, 1, 2, 3, etc) and keeping the projection coordinates as a separate variable, but this doesn't seem to be how CF Conventions wants things.

I've confirmed that tools like panoply can correctly map all the layers, including `land_mask`, `ch4_total` and all sectors. However, do you think this change might make the outputs more trouble to work with?

prayner:
>The most substantial is that the "x" and "y" coordinates are now based on the raw
>projection coordinates in meters. We discussed previously leaving the coordinates as grid
>cell coords (0, 1, 2, 3, etc) and keeping the projection coordinates as a separate
>variable, but this doesn't seem to be how CF Conventions wants things.
>
>I've confirmed that tools like panoply can correctly map all the layers, including
>land_mask, ch4_total and all sectors. However, do you think this change might make the
>outputs more trouble to work with?
Less trouble in my view. Users expect coordinate variables to be in
the unit of the underlying coordinate so this is closer to
expectation. I support the change.

aethr:
The performance regression is in the GFAS layer, and is caused the change to use `xarray` instead of `netCDF4` to load and process the GFAS dataset. I'm going to revert this change since the change to xarray isn't an important part of this PR.

I'd like to see us standardise on using xarray if possible, so it would be good to understand if this performance issue is just an inherent xarray issue, or whether the code could be refactored with xarray to reduce the perf hit.

PR #75, opened 2025-07-01
Output terminology and filename

aethr:
## Description

Throughout the prior codebase the term "output domain" is used to describe the output file, including the default output filename `out-om-domain-info.nc`. Using the term "domain" in this context is a relic of how the prior was built, and doesn't help describe what's actually being produced.

This PR simplifies language around the output, removing the word "domain", so we're left with:
- `OUTPUT_FILENAME` instead of `OUTPUT_DOMAIN`
- `output_file` instead of `output_domain_file`
- `prior_emissions_ds` test fixture instead of `output_domain`

The only functional change is to the default output filename, which previously was `out-om-domain-info.nc`, has been updated to `prior-emissions.nc`. This is more descriptive and aligns with our other downstream outputs such as `posterior-emissions.nc`.

## Checklist

- [x] Tests added

## Notes

Previously, this PR aimed to change the default output filename to a pattern including the domain name and version, like `prior_emissions.{domain_name}.{domain_version}.nc`. However, we agreed after discussion that a static output name will be simpler for downstream consumers, and putting the domain details as attributes in the file makes them easier to work with.

### DIFF
--- a/changelog/75.breaking.md
+++ b/changelog/75.breaking.md
@@ -1,0 +1,2 @@
+Default output filename has been changed from `out-om-domain-info.nc` to
+`prior_emissions.nc`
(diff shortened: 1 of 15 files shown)


prayner:
I agree that using attributes is a better idea than using filenames.

PR #78, opened 2025-07-07
Replace semver domain versioning (v1.0.0) with simple increment (v1 -> v2)

aethr:
## Scope

Although there are circumstances where semver might make sense for domain files, like changes to attributes with no fundamental change to geometry, it seems like adopting semver here adds more complexity than utility. If the domain file changes at all, it's likely that downstream consumers will have to make updates regardless.

In the event that we have a truly breaking change, like a fundamental change to the geometry, we may even consider that a new domain.

Lastly, it seems that semver was either never fully adopted in downstream projects, or was actively walked back, as the infra repo continues to use "v1" everywhere instead of "v1.0.0". This change simply reflects how domains are actually being used.

## Checklist

- [x] Tests added

### DIFF
--- a/changelog/78.breaking.md
+++ b/changelog/78.breaking.md
@@ -1,0 +1,2 @@
+Official Open Methane domains will now follow a simple incremental versioning
+scheme (v1, v2, etc.) rather than semver (v1.0.0)


prayner:
On rereading this it feels slightly odd that the version number is
folded directly into the domain name rather than being appended to it
to construct the full string.

prayner:
my previous concern was only relevant for tests so no problem

prayner (approved):

PR #81, opened 2025-07-11
Grid helper

aethr:
## Description

Add a Grid utility class to provide predictable access to properties of the grid like coordinates and size, as well as common transformations like coordinate lookups.

## Checklist

- [x] Tests added


prayner (approved):

aethr:
> find function in the grid class uses math.floor(a/b) where I'm more used to int(a//b)

I also found that weird, but it was copied from a few places like:
- https://github.com/openmethane/openmethane-prior/blob/main/src/openmethane_prior/layers/omElectricityEmis.py#L63-L65
- https://github.com/openmethane/openmethane-prior/blob/main/src/openmethane_prior/layers/omFugitiveEmis.py#L84-L86

I suppose the result is the same for a single calculation, but using `//` would make it easier to make this work over batches of coordinates? I certainly think that would be useful.

I'll add a test case and see what I can do!

aethr:
Swapping `math.floor` out for `//` was straightforward, no tests broke.

In terms of making it work for batches of coordinates, I think I need to design a test case with you to understand how this would be used before I go ahead and spend time on it. I think we can gradually accumulate more helper functions in this class to move common code out of the layers bit by bit, but it doesn't all need to happen in this PR.

PR #83, opened 2025-07-14
Grid cell name

### LINKED ISSUE #9: Add a unique identifier for each grid cell (geocoding)
## The motivation

A primary feature of the web interface is to be able to reference and visit individual grid cells. By creating a unique identifier via a geocoding scheme for each grid cell, humans will be able to associate an identifier with an important location.

#### Grid cell tiles

When viewing a collection of grid cells, each has an identifier (`N63.326` in this example):

image

#### Viewing a grid cell on the map

When viewing the map, individual grid cells can be selected. The cell identifier is displayed near the top of the left-hand panel (`V32.620` in this example), and we can imagine the URL of the view will include the same identifier, like `openmethane.org/map/X-123-456`:

image

### Why generate in the prior?

**Q: Since the cell id is a user interface tool, why should we generate it in the prior? It could be generated by the web interface when the grid data is ingested.**

A: The prior is responsible for generating the grid itself, so if each cell needs a unique id, this seems like the logical place to generate it. It will also help us correlate user feedback from the web interface, which will use visible identifiers to denote grid cells, with locations in the model, without referencing the database of the web project.

### Identifier requirements

The system to generate identifiers should make ids that are:

- as short as possible
- URL-safe (ie, avoid special characters apart from '-')
- avoid accidental meaning or obscenity
- support 10km grid, but can be extended to 1km or possibly other grid resolutions

## The proposed solution

I would propose a 3 part scheme: `{GRID_ID}.{CELL_COLUMN}.{CELL_ROW}`

- `GRID_ID` - a unique identifier for the grid scheme.
  - For the 10km grid this might have a value of `10`
  - For the 1km grid this might have a value of `1`
  - Although `10` and `1` have a semantic meaning, we could use any non-semantic values for future schemes
- `CELL_COLUMN` and `CELL_ROW`
  - a simple index value from 0 to n where n is the last column/row
    - longitudes in the grid span ~ 6321km so values of `0..633` (10km) or `0..6321` (1km)
    - latitudes in the grid span ~4485km so values of `0..449` (10km) or `0..4485` (1km)
- Examples:
  - `10.0.0` through `10.633.449`
  - `1.0.0` through `1.6321.4485`

**Pros**
- scheme is very simple and easy to understand
- vertical and horizontal adjacent grid cells have similar ids
  - the adjacent cell id in every direction can be guessed
- scheme can be extended to support multiple grid sizes
- some derivation of lat/long can be derived from row/col values

**Cons**
- for 1km grid, ids are quite long, ie `1.6321.4485` is probably longer than desired
- custom scheme is not globally addressable

### Serial ids

In theory, based on the domain, each grid will have a fixed number of cells:
- 10km has approx `283768` cells
- 1km has approx `28357129` cells

We can simply use a serial id for each cell based on the order of generation.

**Pros**
- simple scheme
- adjacent cells have similar ids, but only in 1 dimension

**Cons**
- lack of any semantic meaning
- lookup by index only

### Random ids or uuids

Although this is a standard practice in many databases, it has very few advantages in this usecase. We do not care if ids can be guessed as grid cells are not sensitive data.

This approach has all the disadvantages of serial ids and more. For random values to work without many collisions, the domain must be much larger than the largest possible value, leading to larger-than-necessary ids.

### Encoding `CELL_COLUMN` and `CELL_ROW`

To further shorten and make more ids more memorable we could encode the cell column and row using an encoding scheme that includes alpha characters.

 - base64
   - efficient and URL safe
   - implementations available in every language
   - Cons:
     - can lead to unintended words
     - similarity of `0` and `O` and other characters can be problematic
 - base32 schemes such as:
   - [Geohash](https://en.wikipedia.org/wiki/Base32#Geohash)
   - [Word-safe alphabet](https://en.wikipedia.org/wiki/Base32#Word-safe_alphabet)
     - `10.0.0` becomes `10.2.2` (or `10.22.22` padded)
     - `10.632.449` becomes `10.jX.3P`
     - `1.6321.4485` becomes `1.V78.8J6` (nice and short!)
   - Cons:
     - Less efficient than base64, but unlikely to matter much at the sizes we need
     - Non-standard, so custom implementation is necessary for encoding/decoding
     - Starts at `2` :joy: 

## Additional context

@prayner @lewisjared if this feature is desirable in the prior, I'll leave it up to you to decide where it might be implemented, although I'm happy to attempt a PR if desired.

It would also probably be good to adapt `omGeoJSON.py` or create a new script with the sole intention of outputting the grid itself, without requiring all the inputs and processing to generate the prior. Although we'll only need to do it once, I'd like to have a reproducable way to generate a grid that's consumable by the front-end, as we may need to do this again when we introduce new grid resolutions in the future. Happy for this to be a separate issue/PR however.

aethr:
## Description

Resolves #9.

This PR adds a `cell_name` variable to the prior output, which contains a unique, short name for every cell based on the parent grid slug and the cell coordinates. The function has already been used to generate grid cell names for the front-end, this simply moves it into the back-end and pre-generates the names for later use.

This has the obvious utility that the naming scheme becomes embedded in the output from very early stages, and can be shared by all downstream consumers without re-implementing the encoding. The code in the front-end for generating these names can be removed and they can be read directly from the prior output into the web app database.

The downside is the additional storage this adds to the output file. Testing with and without the `cell_name` variable for a 3-day prior on the "aust10km" domain yields an increase from 21Mb to 31Mb, so roughly 10Mb or 50%. Although the relative increase is large, I don't see 10Mb posing a serious problem for us or downstream consumers.

## Checklist

- [x] Tests added

### DIFF
--- a/changelog/83.improvement.md
+++ b/changelog/83.improvement.md
@@ -1,0 +1,1 @@
+Add "cell_name" to output file with a short unique name for each grid cell
--- a/src/openmethane_prior/cell_name.py
+++ b/src/openmethane_prior/cell_name.py
@@ -1,0 +1,51 @@
+
+alphabet = '0123456789ABCDEFGHJKMNPQRSTVWXYZ'
+
+def encode_word_safe(v: int) -> str:
+    """
+    Encode an integer (like a grid coordinate) as a short string using
+    url-safe characters.
+    """
+    remaining = v
+    output = ''
+    i = 1
+    while i == 1 or remaining > 0:
+        digit = remaining % (len(alphabet) ** i)
+        if digit != 0:
+            remaining -= digit
+        remaining = remaining // (len(alphabet) ** i)
+        i += 1
+        output = f"{alphabet[digit]}{output}"
+    return output
+
+def decode_word_safe(v: str) -> int:
+    """
+    Decode a value encoded with encode_word_safe back to an integer.
+    """
+    output = 0
+    for place in range(len(v), 0, -1):
+        digit = v[place - 1]
+        value = alphabet.index(digit)
+        output += value * (len(alphabet) ** (len(v) - place))
+    return output
+
+def encode_grid_cell_name(grid_name: str, x: int, y: int, separator: str = '.') -> str:
+    """
+    Turn the three components that identify a grid cell (grid, x and y) into a
+    unique, URL-safe string.
+    """
+    return separator.join([grid_name, encode_word_safe(x), encode_word_safe(y)])
+
+def decode_grid_cell_name(grid_cell_name: str, separator: str = '.') -> dict:
+    """
+
+    :param grid_cell_name:
+    :param separator:
+    :return:
+    """
+    grid_name, enc_x, enc_y = grid_cell_name.split(separator)
+    return {
+        'grid': grid_name,
+        'x': decode_word_safe(enc_x),
+        'y': decode_word_safe(enc_y)
+    }
--- a/src/openmethane_prior/outputs.py
+++ b/src/openmethane_prior/outputs.py
@@ -181,6 +181,9 @@ def create_output_dataset(
     prior_ds.time.encoding["units"] = time_encoding
     prior_ds.time_bounds.encoding["units"] = time_encoding
 
+    # compress cell_names which take a lot of space
+    prior_ds.cell_name.encoding["zlib"] = True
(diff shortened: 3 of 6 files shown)


prayner:
Is the size increase with compressed netcdf storage turned on?

PR #84, opened 2025-07-16
Move start and end dates into config

aethr:
## Description

Due to the normalised output format, all layers require a start date and end date so that correctly encoded time steps can be created. By providing these dates in the config object, we can simplify the interface to every layer.

In the current implementation, start and end date can still be read as CLI args, but doing so stores them as environment variables which will be read in to PriorConfig. This should mean no change to downstream users, but there is now the option to provide dates via `START_DATE` and `END_DATE` env variables, which is used by many other openmethane scripts.

## Checklist

- [x] Tests added

### DIFF
--- a/changelog/84.enhancement.md
+++ b/changelog/84.enhancement.md
@@ -1,0 +1,1 @@
+Start date and end date can now be specified via START_DATE and END_DATE environment variables


prayner:
Minor comments only:
I see that most methods are called with the whole config object gut
download_gfas is called explicitly with dates. Should we change for
consistency?

I'm quite impressed that software engineers use the word idempotent,
until last year I hadn't seen it since 2nd year linear algebra

aethr:
> I see that most methods are called with the whole config object but download_gfas is called explicitly with dates. Should we change for consistency?

I considered this, but I felt that `download_GFAS` was a sort of generic method for fetching GFAS, so the dates make a sensible API there. In theory we could move this into `omDownloadInputs` for example, where dates might be specified a different way. It could even be moved out into a separate "library", but I didn't think that was necessary.

prayner:
cool, happy with that
-- 
Peter Rayner (he/him)
My work days are Monday-Thursday, responses at other times might be slow.
The Superpower Institute <https://www.superpowerinstitute.com.au/>
Level 3 105 Victoria St. Fitzroy 3065
Transparent, actionable and auditable data on Australia's methane emissions <https://openmethane.org>
mobile +61 402 752 379
mail-to: ***@***.***

Issue #68, opened 2025-06-05
Modular layer architecture


aethr:
## The motivation

In its current state, running the prior performs a step-by-step imperative process where:
- inputs for all layers are fetched, with some exceptions (GFAS)
- some inputs are transformed into intermediates
- each layer reads its inputs, processes them, and writes its data to the result file

There are several drawbacks:
- hard for fetching to be done separately, which would be beneficial in production since it has a much higher likelihood of failure due to reliance on external systems
- there's no clear relationship between inputs and the layers which use them
- hard to parallelise
- harder for new contributors to understand
- individual layers are harder to test

## The proposed solution

We would like to create a modular architecture where layers are more self-contained, and are invoked by the prior in more predictable ways. For example, each layer would have its own folder in the project, and follow a set of conventions allowing an external caller to:
- fetch its inputs, which results in certain files being written
- invoke its processing step, which returns a data object
- write its processed output to a common file

## Alternatives

Instead of relying on conventions, we could create a strict interface which each layer module must implement, providing all relevant details and callable methods. A central orchestrator could then request, organise and dispatch work that needs to be completed to achieve a result.

At this stage, we don't think the complexity of the prior necessitates this and may make the project as a whole harder to understand for newcomers.

## Additional context

In general, each layer follows a similar pattern:
- fetch some source data
- (sometimes) transform that data into the grid domain
- allocate emissions in the grid domain based on some science or math
- write a layer to the output NetCDF file

There are exceptions to this pattern:
- some layers use the same inputs
- some layers use multiple inputs
- some inputs are transformed into the domain grid, and then used by multiple science layers

If we want layers to be logically separate, but we also want fetching and processing multiple layers to be done in parallel, we need to provide the system with enough information to avoid duplication of work. Ie, if two layers both use "nighttime_lights.TIF", how does the system avoid downloading it twice?

PR #85, opened 2025-07-16
Write output once

aethr:
## Description

Instead of writing output to disk at each stage of the prior (initialisation, each layer, summing the total), this change re-architects the process to build a Dataset from initialisation through layer processing, eventually writing to disk at the end of the process.

## Checklist

- [x] Tests added

### DIFF
--- a/changelog/85.enhancement.md
+++ b/changelog/85.enhancement.md
@@ -1,0 +1,1 @@
+Output will only be written to disk at the end of processing
--- a/scripts/omPrior.py
+++ b/scripts/omPrior.py
@@ -31,7 +31,7 @@ from openmethane_prior.layers import (
     omTermiteEmis,
     omWetlandEmis,
 )
-from openmethane_prior.outputs import sum_sectors, initialise_output
+from openmethane_prior.outputs import add_ch4_total, create_output_dataset, write_output_dataset
 from openmethane_prior.raster import reproject_raster_inputs
 from openmethane_prior.verification import verify_emis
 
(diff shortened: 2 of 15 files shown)


aethr:
Measured running a 3-day prior on `aust10km` before/after this change, and there is minimal impact on performance.

Before:
```
	User time (seconds): 138.60
	System time (seconds): 15.33
	Maximum resident set size (kbytes): 16392732
	Voluntary context switches: 5815
	Involuntary context switches: 1131
	File system inputs: 1158648
	File system outputs: 1739160
```

After:
```	User time (seconds): 134.00
	System time (seconds): 15.54
	Maximum resident set size (kbytes): 16396752
	Voluntary context switches: 5780
	Involuntary context switches: 1685
	File system inputs: 1095616
	File system outputs: 1278880
```

No significant change in overall time (System time) or memory usage (Maximum resident set size). Context switches did seem to increase but this may have been a result of other things happening on the machine.

The major difference is in file system inputs and outputs:

- File system inputs: 1095616 / 1158648 = 94.5% reads
- File system outputs: 1278880 / 1739160  = 73.5% writes

Although this doesn't result in a wall-clock time improvement, it may make a difference on a more resource constrained environment.

aethr:
Seeing how similar every `__main__` entrypoint is in individual layer scripts does tell me that there is an abstraction missing here. It seems we want a way to run the prior, while selecting which layers to run or not run. But every other part of the process is always desired:
1. parse env and cli args into config
2. prepare output dataset
3. run one or more layers
4. sum sector layers into a total
5. write output

There's a step missing from this which is "download inputs", which is currently done in a separate script, and downloads inputs for all layers, except GFAS which is downloaded on demand.

Figuring out how to make step 3 easy to control, and possibly how to fold in "download inputs" but only for layers being acted on, is the problem to solve in #68.

prayner:
agreed. Abstracting the body of the __main__ methods need not break
the ability to run each layer separately for testing.

aethr:
@prayner confirming that the behaviour after this PR is that each time any layer is run, the entire file will be re-created, not appended.

It wouldn't be hard to add an "append" mode which simply replaces:
```python
    ds = create_output_dataset(config)
```
with something like:
```python
  ds = xr.load_dataset(config.output_file)
```

prayner:
wouldn't bother with the extra arg yet, just good to document the
change in behaviour in this thread

PR #42, opened 2024-08-06
fixing projection and land maskingLocation fix

prayner:
## Description
The previous code was making some wrong assumptions about the domain projection so various emissions were displaced, especially in smaller domains. These changes fix this and remove some of the intermediate reprojections. Also applies land masking to all fields except fugitives to remove offshore emissions arising from coarse grids

responds to issues #17 #24 and #31 
## Checklist

Please confirm that this pull request has done the following:

- [ ] Tests added
- [ ] there is now a reprojection test which picks a single raster point and tests it projects to one point and that its location is approximately unchanged
- [ ] Documentation added (where applicable)
 

## Notes
there is still one landuse intermediate projection which we should also remove

PR #87, opened 2025-07-22
Normalise approach to reprojecting coordinates

aethr:
## Description

Having a well tested, centralised capability for reprojecting coordinates into the domain grid, this approach should be normalised and adopted wherever possible in the project.

This is equivalent to the initial series of commits in #42, but based on recent additions to the codebase.

## Checklist

- [x] Tests added

### DIFF
--- a/changelog/87.improvement.md
+++ b/changelog/87.improvement.md
@@ -1,0 +1,1 @@
+Adopt standard reprojection approach using Grid class throughout the project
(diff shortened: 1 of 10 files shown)


prayner:
I suggest raising an issue to separate the four layers calculated in
omAgLulucfWasteEmis.py into separate modules. Some of that will happen
anyway when we redo the waste but I still think it's worth getting as
regular a structure as we reasonably can. Not under this PR though.

prayner:
my previous comment also holds for omIndustrialStationaryTransportEmis.py

prayner (approved):

PR #88, opened 2025-07-22
Add optional land masking

### LINKED ISSUE #44: Check for offshore electricity emissions
## The problem
Decide if the `electricity` layer should be land-masked or not.

https://github.com/openmethane/openmethane-prior/pull/43#issuecomment-2285336186

## Definition of "done"
* [ ] Decide if the `electricity` layer should be land-masked or not.

aethr:
## Description

Mask some layers using the land mask where emissions are unlikely to occur. This is done in two ways:
- land masking the input data set
- land masking the result after regridding

Both approaches are used based on whether we want to spatialise an existing inventory total (mask the input before spatialising the inventory) or whether we're regridding an existing, coarse dataset (mask the result after regridding).

Resolves #44.

This is equivalent to part of #42, but rebased onto more recent changes in the codebase.

Note: this PR does not include our proposed masking strategy for agriculture, lulucf and waste emissions, where we propose to mask the input dataset prior to spatialising the inventory totals. The existing implementation counts cells with each land use type before regridding, which makes applying the land mask more difficult. This will be addressed in a follow up PR.

## Checklist

- [x] Tests added

### DIFF
--- a/changelog/88.enhancement.md
+++ b/changelog/88.enhancement.md
@@ -1,0 +1,1 @@
+Apply land mask to emissions sources that should only occur on land
--- a/docs/masking.md
+++ b/docs/masking.md
@@ -1,0 +1,58 @@
+
+# Masking data
+
+In data science, masking is the process of subsetting or excluding spatial data
+based on underlying properties.
+
+The most common type of masking in Open Methane is applying a "land mask",
+which is a binary field that describes whether each grid cell covers "land" or
+"sea". In reality, many grid cells may contain both land and sea, but will the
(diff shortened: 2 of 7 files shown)


prayner:
should we consider documenting in the output file whether a field has
been masked and when?

prayner (approved):

aethr:
> should we consider documenting in the output file whether a field has been masked and when?

I was thinking something along the same lines. Maybe we can introduce this annotation when we implement inventory masking?

PR #90, opened 2025-07-23
Fix omAgLulucfWasteEmis.py reprojection

aethr:
## Description

Previously, the omAgLulucfWasteEmis.py module was using an intermediate reprojected version of the Australian land use file, which was generated using the wrong projection.

This PR adapts the `remap_raster` utility to allow an alternative input CRS to be specified, so that we can use the utility to regrid the land use raster onto the domain grid. The result then has the land mask applied so that all sectoral emissions are applied within valid land pixels.

Based on commits originally authored by @prayner in #42.

## Checklist

- [x] Tests added

### DIFF
--- a/changelog/90.fix.md
+++ b/changelog/90.fix.md
@@ -1,0 +1,1 @@
+Fix reprojection in agriculture, LULUCF and waste layers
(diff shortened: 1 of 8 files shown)


prayner:
Should we also switch over prints to loggers here, e.g. omAgLulucfWasteEmis.py:107

aethr:
@prayner I've been thinking about this, but currently the prior doesn't use loggers at all. My idea was to create a branch which replaces all the `print` with `logger` in one hit across the codebase. What do you think?

prayner (approved):

prayner:
agreed, much cleaner

PR #100, opened 2025-07-31
Add `logger` util to replace `print` statements

aethr:
## Description

This brings the `logger` util from openmethane to provide consistent handling of console output based on the `LOG_LEVEL` env var.

All `print()` statements have been replaced with `logger.info` or `logger.debug` based on whether a user would expect to see that input during normal operation, or whether it's more useful during development.

Some output that doesn't come from `print` (such as dumping the config to screen) is now conditional based on the log level.

## Checklist

- [ ] Tests added

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,28 @@ and end date as below):
 poetry run python src/openmethane_prior/layers/omWetlandEmis.py --start-date 2022-07-01 --end-date 2022-07-01
 ```
 
+### Console output
+
+The detail of console output can be controlled by setting the `LOG_LEVEL` env
+variable. By default, this is set to `INFO`, but more or less can be achieved
+by setting other log levels:
+
+```shell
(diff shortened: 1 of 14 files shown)


prayner:
There are a few places like inputs.py:69 where I would
have info not debug as the level since I think I'd like to know if
it's not downloading things
inputs.py:106 is warning the right level to attach to a sys.exit?

prayner:
>Personally I was surprised that this exit didn't just raise an error. I think printing
>multi-line error output is a bit of a bad practice, and the warning here is actually
>instructions on how to resolve the issue.
>
>I would be inclined to leave the multi-line output as warning and raise an error with a
>terse error message instead of sys.exit(1). That way if a run has limited output to ERROR
>only, they will still see the error in the console, but a log file doesn't end up with
>instructions on how to download files.
agreed.

Issue #55, opened 2024-11-14
incorrect scaling of emissions by spatial proxy leads to failing tests


prayner:
## The problem
Versions of the prior which scale a national emission by a spatial
proxy are failing because the spatial sum of the proxy over the domain
does not match the denominator in the scaling equation local-emission
= national-emission  times proxy /proxy.sum() where proxy is a 2d field.
This is wrong but not simply a code bug. In the first place we should
not expect it to work for domains less than national. That is dealt
with by the CI tests running on the national domain. But secondly
because of the different projections of the proxy and the OM domain
there are always significant parts of the proxy lying outside the
OM domain. There may also be parts of the OM domain lying outside
Australia (e.g. parts of PNG and Indonesia) and these should not be
included. Thus we will need to integrate any proxy over an entity we
can define as Australia. We will need this even for subdomains since
the above spatial-emission equation needs proxy.sum() for Australia,
not tjust the chosen domain.

## Definition of "done"
Tests are passed and the spatial-emission equation is calculated
correctly for national and subdomains. 
We will need:
- some way of defining whether a given point in a proxy lies inside or
 outside Australia
- the inclusion of a national sum of proxies whether or not we are
  using the national domain.
## Additional context
A possible route is the landuse data file already provided since it is
naturally  limited to Australia. It is an inefficient way of
calculating inclusion though, especially at the high resolution of
some of the proxies.

aethr:
@prayner I think I understand the problem.

Perhaps we need to understand how the "national emissions" we're referencing determine the boundaries or inclusions in their data set. Let's discuss in person next week.

prayner:
A little more detail on the solution: 
The key problem is how we calculate the sum of a proxy over a domain
which is (1) consistent with the national inventory, i.e Australia and
(2) consistent with the land mask if it is a masked field. Here is a
proposed algorithm:
- turn the config.layer_inputs.land_use_path tif image into a mask, 1 for land, 0 for ocean
- Use remap_raster to map it to the domain
- Normalise this field by dividing by its maximum value. This should generate something close to a land mask but limited to Australia
- Multiply this by the land mask so it's consistent with how we will treat fields later
- Set all values > 0 to 1 so it is a true mask
- OPtionally store this so we don't need to regenerate it. This should be in the intermediates directory
- Now when proxying any field, either for the national domain or a subdomain:
- Remap the proxy to the national domain 
- Multiply by this Australian land mask.
[...]

PR #103, opened 2025-08-02
Add inventory mask

prayner:
## Description
Partly responds to #55. There it was noted we needed a mask that
accounted for both presence of land and the extent of the national
domain. This achieved by and-ing the LWMASK from WRF and the landuse
data input file with all  nonzero classes considered as Australian
land. This INVENTORYMASK is saved for subsequent use in the prior.
## Checklist

- [ ] Tests added

## Notes
The INVENTORYMASK field should be checked visually. The easiest way is
probably to plot LANDMASK -INVENTORYMASK which should correspond to
land points outside Australia.

move create_domain_info from openmethane repo and add calculation of inventory mask

### DIFF
--- a/scripts/create_prior_domain.py
+++ b/scripts/create_prior_domain.py
@@ -1,0 +1,271 @@
+#!/usr/bin/env python
+#
+# Copyright 2023 The Superpower Institute Ltd.
+#
+# This file is part of OpenMethane.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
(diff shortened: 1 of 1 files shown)


aethr:
The new mask makes sense, although it's a shame we can't cast the rioxarray to bool or uint.

However, I don't think we'll be able to use PriorConfig in this file, particularly `domain_grid`, since that property reads the input domain to determine the grid parameters. What we could probably do is replace `config.domain_grid()` with a new instance of `DomainGrid`, since we've just assembled all the properties we'll need in `domain_ds`:

```python
domain_grid = DomainGrid(domain_ds)
```

The places where we use `config.as_input_file()` could be replaced with the `landuse` arg which we don't appear to be using.

aethr:
I've run this locally with aust10km geometry and MCIP inputs and the INVENTORYMASK appears to be very similar to LANDMASK, but without the land portions of Papua New Guinea. I mapped this along with an Australia shape and it appears to adhere very well to the Australian coastal boundary.

Out of curiosity, I also mapped a logical XOR of the LANDMASK and INVENTORYMASK to see whether any Australian coastal pixels might be excluded, but from a cursory review it appears that only Papua New Guinea pixels have been culled. This looks like it will suit our purposes very well.

prayner:
Ok, sounds good enough to move on. A note that I should update logging
there to match #100

aethr:
@prayner I'll update logging since I'll be making quite a few changes to this file to adopt the new formatting.

Issue #102, opened 2025-08-01
Testing openmethane prior fails for Transport, Stationary, Industrial, Agriculture, and LULUCF for downscaling.


Ida-droid:
## The problem
I'm running 2km spatial resolution, and testing if the downscaling implemented in openmethane-prior works. 
I reran openmethane-prior with my 2k grid and I got a weird percentage distribution of the individual sectors. 

Percentage contributions: (Calculation: sector_i [kg/m2/s] / sum_sectors [kg/m2/s]) = Result given in [%])
Agriculture: 6.55
LULUCF: 19.7
Waste: 18.3
Livestock: 17.0
Industrial: 0.11
Stationary: 2.04
Fugitive: 17.02
Termite: 1.64
Fire: 2.79
Wetlands: 14.32
Electricity: 0.10

I tried to explain what was happening, looking to see if this was already the case for the old 10km spatial grid run version. Old version means the main branch 3 months ago. 
Therefore, I'm trying to define the region in the 10k similarly to the 2k region. 
The sum of the emissions of these regions should be the same. 

## First problem: 
It was hard for me to find the correct i,j index for my inner domain. Even though they are given in the namelist.wps, they do not match with the shape of the openmethane-prior emission.nc.
namelist.wps: 
 i_parent_start    = 1, 350,
 j_parent_start    = 1,120,
 e_we          =467,411,
 e_sn          = 443,391, 
This means that the 10k grid should have a 467 x 443 shape. 
And the nested domain 2k grid should have the shape of 411 x 391. 

However, 
emis_10k.variables['LAT'].shape= 430 x 454
emis_2k.variables['lat'].shape= 378 x 398 

I don't know why these shapes are different. -> Problem unresolved! 

I could still match them by eye, but it is not ideal. 

<img width="636" height="554" alt="Image" src="https://github.com/user-attachments/assets/1ba127c1-3a94-425f-84fd-ca7369a3ac3b" />

<img width="636" height="554" alt="Image" src="https://github.com/user-attachments/assets/ed6ad7b9-1211-4674-89e1-2718a7eb6755" />

After doing so, I multiplied the sum by the length and width of one grid cell. 

Example:
print(np.sum(emis_10k.variables['OCH4_TERMITE'][0,0,119-5:119+78-5,349:349+82])*10000*10000)
print(np.sum(emis.variables['ch4_sector_termite'][0,0,:,:])*2000*2000)

## Second Problem
Results show: Units [kg/s]
Termites
[...]

PR #105, opened 2025-08-08
Use Inventory Mask

prayner:
## Description
this code uses the inventory mask created in #103 to mask relevant
sectors.Responds to #102 and should also allow correct sub domains in
the prior
## Checklist

- [ ] Tests added

## Notes


aethr (commented):
@prayner logic looks good to me.

Let me know if you'd like a hand resolving the merge conflict or getting any tests fixed up.

prayner:
There's still a test for the oCH4_total though the field is
deprecated. Will the CH4_total catch that already?

aethr:
> There's still a test for the oCH4_total though the field is deprecated. Will the CH4_total catch that already?

That test should have a check for ch4_total and OCH4_TOTAL (which have identical values), since it finds all numeric variables and calculates the mean automatically. When we remove OCH4_TOTAL from the output (probably a v1.1.0 sort of change) then we'll get rid of that check as well as LANDMASK I imagine.

PR #107, opened 2025-08-10
make creation of inventorymask dependent on landuse argument

aethr:
## Description

Note: this was intended by @prayner to be included in #103 , but I merged his PR a bit prematurely. Sorry Peter!

The newly created inventory mask will typically only related to the top-level domain, since it represents the area which is covered by national inventory statistics which the sectoral emissions totals are taken from. When running the prior on a subdomain, the input domain will be the smaller subdomain, but the inventory domain will still be the one specified in the top-level domain.

For this reason, the user should be able to create a domain without an inventory mask, since it may not make sense for smaller domains.

## Checklist

- [ ] Tests added

### DIFF
--- a/scripts/create_prior_domain.py
+++ b/scripts/create_prior_domain.py
@@ -97,39 +97,40 @@ def create_domain_info(
         for var in ["LATD", "LOND"]:
             domain_ds[var] = dotXr[var].rename({"COL": "COL_D", "ROW": "ROW_D"})
 
-    print("Loading land use data")
-                # this seems to need two approaches since rioxarray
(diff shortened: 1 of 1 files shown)


aethr:
@prayner I opened this PR when I saw you'd pushed the change to github. I might wait until our other inventory-related PRs have been merged and I'll re-base this so it's up to date.

My only question on this is, is it necessary to have `INVENTORY_MASK` in the domain file at all? If no `landuse_file` is supplied, can we simply omit the `INVENTORY_MASK` attribute from the output file?

In the case of a subdomain, for example, they should be supplying the aust10km domain file as the inventory domain anyway, right?

PR #108, opened 2025-08-13
Fix incorrectly apply inventory mask to different sized domain

aethr:
## Description

We want to apply the inventory mask to spatialised datasets before scaling, so we don't allocate emissions to areas outside of the inventory. However, to apply the mask we need to make it the same shape and cover the same area as the domain. The previous implementation only worked when the inventory and the domain were already on the same grid.

## Checklist

- [x] Tests added

### DIFF
--- a/src/openmethane_prior/grid/regrid.py
+++ b/src/openmethane_prior/grid/regrid.py
@@ -1,0 +1,49 @@
+#
+# Copyright 2025 The Superpower Institute Ltd.
+#
+# This file is part of Open Methane.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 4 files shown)


prayner:
can/should we check equality on the projections?

aethr:
> can/should we check equality on the projections?

@prayner do you mean, if the grids are equal, don't waste time "regridding"?

prayner:
why the float on 76 in omIndustrial... when it's not needed at 140 in
omAg..., xarrays again?

prayner (approved):

prayner:
>@prayner do you mean, if the grids are equal, don't waste time "regridding"?
No, I mean throw an error if they don't share a center etc

aethr:
> why the float on 76 in omIndustrial... when it's not needed at 140 in omAg..., xarrays again?

Precisely!

aethr:
> No, I mean throw an error if they don't share a center etc

Good idea, added.

PR #119, opened 2025-09-01
Add ANZSIC sector codes

### LINKED ISSUE #96: Adopt ANZSIC or IPCCC sector codes
## The motivation

There are sectoral classification schemes which exist to describe sectors of industry that are responsible for emissions, including ANZSIC in Australia, EDGAR codes, and IPCC 2019.

If each layer of the prior was annotated with an ANZSIC code, it would make it more straightforward to extract details like the sector emissions from the national inventory, or allocate SGA facility emissions to specific sectors.

ANZSIC is a hierarchical scheme, so even our existing breakdown between livestock and agriculture could be well modelled by annotating these layers like:
- livestock: 014
- agriculture: 01

## Alternatives

There are other classification schemes, such as EDGAR codes or IPCC 2019 sector identifiers. A mapping of some of these can be viewed in Table 10 "Mapping emission sources to sectors" of the [IPCC Annex II: Definitions, Units and Conventions](https://www.ipcc.ch/report/ar6/wg3/downloads/report/IPCC_AR6_WGIII_Annex-II.pdf).

Using a global scheme would make it easier to develop a global prior or utilise global datasets. However, we would need to find or maintain a mapping between any global scheme and ANZSIC to be able to align this with Australian official data sources.

aethr:
## Description

ANZSIC codes are a foundational reference that the Australian government uses to report on emissions. All sectoral emissions data released by the government uses the ANZSIC taxonomy to segregate different sectors.

This makes ANZSIC useful to the prior for places where we seek to use government data as an input (sectoral emissions from NGGI, or Safeguard Mechanism data), and it also serves to help us compare Open Methane emissions estimates back to government reporting.

Resolves #96.

## Checklist

- [ ] Tests added

### DIFF
--- a/src/openmethane_prior/layers/omAgLulucfWasteEmis.py
+++ b/src/openmethane_prior/layers/omAgLulucfWasteEmis.py
@@ -43,20 +43,24 @@ logger = logger.get_logger(__name__)
 sector_meta_map = {
     "agriculture": SectorMeta(
         name="agriculture",
+        natural=False,
         cf_standard_name="agricultural_production",
     ),
     "lulucf": SectorMeta(
         name="lulucf",
+        natural=False,
         cf_standard_name="anthropogenic_land_use_change",
     ),
     "waste": SectorMeta(
         name="waste",
+        natural=False,
         cf_standard_name="waste_treatment_and_disposal",
     ),
 }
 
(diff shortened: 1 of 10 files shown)


aethr:
Closing in favour of #121 .

PR #122, opened 2025-09-10
Use Australia National Greenhouse Accounts inventory

### LINKED ISSUE #73: Use inventory data directly from official sources
## The motivation

Currently, the prior uses a bespoke `ch4-sectoral-emissions.csv` file as the input for sectoral totals which are then spatialised in various layers. This file includes numbers that have been calculated from NGGI data, but it's not clear from the code what algorithm was used, or what time period these numbers represent.

The Australian government provides greenhouse gas inventory data in a number of formats (see below), by using one of the official sources directly we gain:
- accountability for how sectoral totals are derived
- ability to use the data for the quarter or year currently being evaluated by the prior (or the most recent available)
- ability to easily adopt the newest quarterly NGGI data when it becomes available

### Official sources

The government provides inventory in a number of ways:

#### National Greenhouse Accounts - https://greenhouseaccounts.climatechange.gov.au/

Poorly built web app which provides inventory data in a few formats. Navigating to Emissions Inventories gives access to interactive tools which provide yearly methane emissions disaggregated by:
- location (state)
- gas(es), including CH4
- sector(s)
- time period (from 1990, two year lag - 2023 figures published May 2025)

Results are charted, but can be downloaded as CSV. Depending on which inventory tool is selected, sectors can be chosen based on:
- ANZSIC category ("National inventory by economic sector")
- UNFCCC CRT category ("Paris Agreement inventory")

The app also provides access to an API for "Bulk download" in various formats. The API appears to be "OData", although it doesn't support any of the useful OData features (like filtering). The "AR5_ParisInventory_AUSTRALIA" endpoint appears to be the most useful to us, which provides emissions by year, gas and sector, using UNFCCC CRT categories.

```
# bulk download
curl -s https://greenhouseaccounts.climatechange.gov.au/OData/AR5_ParisInventory_AUSTRALIA > data/AR5_ParisInventory_AUSTRALIA.json`

# filter to only CH4 emissions
curl -s https://greenhouseaccounts.climatechange.gov.au/OData/AR5_ParisInventory_AUSTRALIA \
  | jq '.value[] | select(.Gas_Level_0 == "CH4")'
```

Returned data looks like:
```json
{
  "@odata.context": "https://greenhouseaccounts.climatechange.gov.au/odata/$metadata#AR5_ParisInventory_AUSTRALIA",
  "value": [
    {
      "ID": 155615542418,
      "InventoryYear_ID": 2023,
      "UNFCCC_Level_0": "Total UNFCCC",
      "UNFCCC_Level_1": "Land Use, Land-Use Change and Forestry UNFCCC",
      "UNFCCC_Level_2": "Wetland",
      "UNFCCC_Level_3": "Land converted to Wetland",
      "UNFCCC_Level_4": "Land converted to Flooded lands",
      "UNFCCC_Level_5": "Reservoir Methane",
      "UNFCCC_Level_6": null,
      "UNFCCC_Level_7": null,
      "UNFCCC_Level_8": null,
      "UNFCCC_Level_9": null,
      "UNFCCC_Level_10": null,
      "Gas_Level_0": "CH4",
      "Gas_Level_1": null,
      "Gas_Level_2": null,
      "Gas_Level_3": null,
      "Gg": 0.5553692029907528
    }
  ]
}
```

This is the same data available in the web app, so has the same time periods, etc. While this endpoint doesn't allow disaggregation by state, the API listing shows that endpoints are available for each Australian state and territory.

One potential issue with this data is the UNFCCC categories are only referenced by name and not by code. We may need to develop a mapping between the names used in this API and the codes that we intend to use for layer classification in #96.

#### NGGI quarterly reports - https://www.dcceew.gov.au/climate-change/publications/national-greenhouse-gas-inventory-quarterly-updates

The quarterly reports are documents which report and explain the latest inventory numbers, released every 3 months with up to 6 month lag (June 2025 data published 27 August 2025, Dec 2024 data published 30 May 2025) . In the reports are numerous charts which break down emissions in several ways (ie by gas, location, sector, etc). The data used to create these charts is made available as Excel spreadsheets.
  
The data is harder to use as there's no direct accounting of methane gas by sector, but using some heuristics we may be able to derive reasonable conclusions based on the data.

## The proposed solution

- Fetch a copy of "Bulk API" data and store it in S3 with prior inputs
  - document exactly how the file was fetched / processed for auditability
- Create a mapping between UNFCCC codes in the bulk data and UNFCCC codes added in #96 
- Create an abstraction in the codebase to get the inventory emissions within a start/end date for one or more IPCC sectors
  - Use the Bulk API input as a data source
  - Filter / aggregate based on time period and provided UNFCCC sectors mapped to inventory sectors
- Replace any figures currently based on `ch4-sectoral-emissions.csv` with this abstraction

## Alternatives

If we use the Bulk API data as our primary source, we will only have access to data that is 2 years behind. If we wanted to generate a prior result for a period after the available data we could:
- use the latest figures we have
  - could be off by a significant percentage depending on the sector
- extrapolate trends in the data forward
  - could be dangerous and we would need to invent a methodology
- supplement the annual data which is broken down in detail, with more recent high level figures from NGGI quarterly updates
  - #120 

## Additional context

This should help to resolve questions raised by @Ida-droid about how the values in `ch4-sectoral-totals.csv` were calculated.

aethr:
## Description

Resolves #73.

Replaces the undocumented `ch4-sectoral-emissions.csv` input file with data from Australian National Greenhouse Accounts "Paris Agreement" inventory, broken down by UNFCCC sector.

The work consistutes two major parts:
- map from inventory UNFCCC category names to category codes
- search through the inventory for emissions matching a list of sectors to compute a sector total

## Checklist

- [ ] Tests added


aethr:
I've constructed a test that is meant to check new inventory totals (using 2023 as a reference year) against the previous totals from `ch4-sectoral-emissions.csv`. Most of the sectors look very close, but `agriculture` and `fugitive` are alarmingly different.

```
fugitive: 0.0429 from 1.1739
agriculture: 0.2187 from 2.2825
electricity: 0.0111 from 0.0143
stationary: 0.0466 from 0.0529
waste: 0.4783 from 0.4729
transport: 0.0115 from 0.0114
industrial: 0.0029 from 0.0029
lulucf: 0.4869 from 0.51
```

prayner:
Agreed, the only one which surprises me is lulucf. fwiw the most
useful form for these numbers for me is to multiply that mean by the
number of grid cells and the cell_area then divide by 1e3 to get
tonnes of methane. Also useful to take the absolute difference which
is the number I focus on most. But, in general, looking good I think.

aethr:
> sector/inventory.py:47 seems some confusion between kilotonne and gigatonne

Fixed in [81b26ce](https://github.com/openmethane/openmethane-prior/pull/122/commits/81b26ce60812d397d9d032f7fbfec143708a7d33).

> sector/inventory.py:69 float(str) seems to work even on scientific notation

Fixed in [46e8673](https://github.com/openmethane/openmethane-prior/pull/122/commits/46e8673cffdcab8a327e843e8e3bb5ed2bec0913).

prayner (approved):

Issue #91, opened 2025-07-23
Separate each sector implementation into its own module


aethr:
## The motivation

Relates to #68.

Although a few sectors currently share a data source and a methodology, the codebase may be easier to understand if each sector was separated into its own python module.

Modules which currently add multiple sectors:
- [layers/omAgLulucfWasteEmis.py](https://github.com/openmethane/openmethane-prior/blob/main/src/openmethane_prior/layers/omAgLulucfWasteEmis.py)
  - livestock
  - agriculture
  - LULUCF
  - waste
- [layers/omIndustrialStationaryTransportEmis.py](https://github.com/openmethane/openmethane-prior/blob/main/src/openmethane_prior/layers/omIndustrialStationaryTransportEmis.py)
  - industrial
  - stationary
  - transport

Each sector module would export an object which implements a common interface, defining its meta attributes, data sources and the method for creating the emissions estimate. This will make it easier to create a single point of entry which can work across all sectors, enabling things like:
- creating the prior output
- fetching all data sources
- determining sector coverage

Using a common entrypoint, the user can select which layers to operate on, since every layer will have a defined name, making something like this possible:
```shell
poetry run python scripts/omPrior.py --sectors agriculture,lulucf,fires
```

This command would only download data sources and process emissions for the selected sectors.

## The proposed solution

Create a Python module for each sector. Export a consistent interface from the module that can be called by the parent script.

### SectorEmissions

Each sector will implement a SectorEmissions interface, which will contain:
- the `SectorMeta` attributes
- `processEmissions` method (probably renamed `process_emissions`)
- `data_sources`

Instead of a subclass, this will probably be implemented as a parameterised object, like:
```python
sector = SectorEmissions(
    name="fugitive",
    emission_category="anthropogenic",
    unfccc_categories=["1.B"], # Fugitive emissions from fuels
    cf_standard_name="extraction_production_and_transport_of_fuel",
    process_emissions=process_fugitive_emissions, # defined elsewhere
    data_sources=[
        coal_facilities_data_source,
        oil_gas_facilities_data_source,
    ],
)
```

### Project structure

Currently, the openmethane_prior package mixes general utilities and behaviors with the sector implementations. As part of this refactor, I propose defining three top-level folders under `openmethane_prior`:

- `lib` - general utilities and abstract code, ie:
  - `lib/config`
  - `lib/data_manager`
  - `lib/sector`
- `data_sources` - shared or generic DataSource implementations
  - `data_sources/inventory`
[...]

PR #135, opened 2025-10-15
Separate all sectors into independent modules

### LINKED ISSUE #68: Modular layer architecture

aethr:
## Description

Resolves #68, #91.

This change removes the ability to run a sector implementation as a main script, but adds the `--sectors` argument to `scripts/omPrior.py` so that a single or subset of sectors can be processed.

```shell
# No longer works
poetry run python src/openmethane_prior/layers/omAgLulucfWasteEmis.py

# Works
poetry run python scripts/omPrior.py --sectors agriculture,livestock,lulucf,waste
```

## Checklist

- [ ] Tests added


prayner:
scripts/omPrior.py:44: this is right but putting all the testing in
the list comprehension meant I had to read it three times,
sectors =list(all_sectors) if config.sectors is None else [s for s in
list(config.sectors)]
seems cleaner
config.py:221 and beyond: why are we putting these values into the
environment or am I misunderstanding something?

lib/prior.py: the name of this file seems a bit generic, I also wonder
if it's in the right place.
lib/prior.py:55: what happens to sector_data? If it's unused and you
want to make that clear the convention seems to be to call the output
_. Also worth noting that the create_estimate methods must call
add_sector in the current implementation. I think we discussed why
that was required but it seems odd to me now, creating an estimate and
adding it to a prior seem different operations.
add_sector?
sectors/__init__.py but really a more general point. Probably should
wite a cheatsheet on adding a new sector e.g. knowing you have to
modify the all_sectors list by hand.

I confess I skimmed the changes that were mainly structural.

prayner (approved):

prayner:
>I agree about the naming, I'm very open to suggestions.
I quite like run_prior since it's the main method in there and that's
pretty much what it does. 
>I do think this belongs in src/lib as opposed to scripts.
agree about src rather than scripts. I think anything in scripts
should be runnable from the cli.  "lib" has a feeling of subsidiarity about it rather
than being the major entry point, not sure it's worth a new folder
though. What about moving it up one level  to src? untidy?
> lib/prior.py:55: what happens to sector_data? If it's unused and you want to make that
> clear the convention seems to be to call the output. Also worth noting that the
> create_estimate methods must call add_sector in the current implementation. I think we
> discussed why that was required but it seems odd to me now, creating an estimate and
> adding it to a prior seem different operations.
>
>I agree that adding the sector data to the output should probably be handled in run_prior.
>We have everything we need to do it cleanly and reduce the amount of duplicated code needed
>in sectors. We had discussed that it wasn't strictly necessary to get this PR over the
>line, but I would be happy to add it, otherwise it might never get
>done.
Does seem a sensible part of the restructure already in #135 so
I recommend adding it.
>
> Probably should write a cheatsheet on adding a new sector e.g. knowing you have to
> modify the all_sectors list by hand.
>
>I'll add something to the README after I make the update above to run_prior.

aethr:
> I quite like run_prior since it's the main method in there and that's pretty much what it does.

Probably being pedantic, but "run" still implies this is a runnable script. What about "create_prior"?

> "lib" has a feeling of subsidiarity about it rather than being the major entry point, not sure it's worth a new folder though. What about moving it up one level  to src? untidy?

In a way, this is just another function. I do think if we rename the file "run_prior" or "create_prior" it will be a more obvious point of importance. I also think that many of the files in `lib/` could be moved into a `lib/util/` folder, which would give really important things in `lib` more prominence.

My suggestion would be:
- rename `prior.py` to `create_prior.py`, and rename `run_prior` function to `create_prior`
- instead of writing to the output file in `create_prior`, return a DataSet instead
- move the last two lines of `create_prior` (verification and writing output) back into `omPrior.py`
- move some of the utilities (like logger, raster, units, utils) into a `lib/util` folder

prayner:
>Probably being pedantic, but "run" still implies this is a runnable script. What about
>"create_prior"?
>* rename prior.py to create_prior.py, and rename run_prior function
to create_prior
agreed
>* instead of writing to the output file in create_prior, return a DataSet instead
>* move the last two lines of create_prior (verification and writing output) back into
> omPrior.py
agreed.
>* move some of the utilities (like logger, raster, units, utils) into a lib/util folder
Don't see the need for that level of hierarchy, e.g. I couldn't think
of clear criteria about what goes where. But go ahead if it makes the
layout clearer.

Issue #94, opened 2025-07-30
Use emissions reported under the Safeguard Mechanism (SGA)


aethr:
## The motivation

Australia has introduced legislation which requires large emitters to report their emissions for every facility which emits [more than 100,000 tonnes of carbon dioxide equivalent (CO2-e) per year](https://www.dcceew.gov.au/climate-change/emissions-reporting/national-greenhouse-energy-reporting-scheme/safeguard-mechanism#toc_1). These facilities cover multiple sectors, but are mostly in the mining, oil & gas, and waste sectors.

The Australian Clean Energy Regulator should publish a dataset each year, which includes the total methane emitted for that year, from all facilities covered by the Safeguard Mechanism. The 2023-2024 dataset can be downloaded here, by expanding the "About this table" section: https://cer.gov.au/markets/reports-and-data/safeguard-data/2023-24-baselines-and-emissions-data#baselines-and-emissions-table

Within this CSV, useful columns include:
- Facility name
- State/Territory of operation
- ANZSIC - the sectoral classification this facility falls under, ie "Coal mining (060)"
- GHG Methane, assuming this is tonnes of CO2-e, so may need to be converted into kg/year CH4

This CSV does **not** include:
- a unique dentifier
- a location

This leaves us with two problems to solve:
- how do we allocate these reported emissions to a location
- how do we ensure the same location does not have emissions allocated in other parts of the prior?

## The proposed solution

TBC

Ida-droid:
Safeguard data exists for the financial year 2023-2024. 
If you want to go back in time, you might wanna look at https://www.nature.com/articles/s41597-023-02275-z

aethr:
It does appear from [this documentation](https://cer.gov.au/system-content/eers/help/22-23/facilityentity-facilityentity-edit) that EERS (the system used by corporations to report emissions to the Clean Energy Regulator) requires a street address, lat/lon and ANZSIC classification for **every facility**.

It would be incredibly helpful to get access to the data if it's public.

aethr:
A brief analysis of the SGM 2023-24 Baselines and Emissions table:
- Total facilities: 237
- Facilities reporting more than 100t CH4: 101

Sectoral breakdown of the 101 largest methane emitters:
- 61 Coal mining (060)
- 27 Oil and gas extraction (070)
- 5 Gas supply (270)
- 3 Pipeline and other transport (502)
- 2 Petroleum and coal product manufacturing (170)
[...]

PR #139, opened 2025-10-27
Add Safeguard Mechanism data source

aethr:
## Description

This is a step towards #94.

This PR adds the 2023-24 Safeguard Mechanism Baselines and Emissions Table dataset as a new data source, but does not use the data in any sector yet.

It includes a helper utility to find SGM facilities by ANZSIC sector, which will be the first step towards including SGM facilities in any sector implementation.

## Checklist

- [x] Tests added

## Next steps

Follow on PRs will be added to:
- split "fugitive" sector implementation into coal and gas (#140) 
- add ANZSIC code(s) as a `PriorSector` attribute
- add a supporting data source with locations of SGM facilities
- use SGM facility emissions in the coal layer

## Notes

Worth noting that this PR makes minimal attempt to allow for Safeguard estimates before or after the current single dataset we have available, and explicitly does things like:

```
        ch4_emissions=dict({
            "2023-2024": record.co2e_ch4 / ch4_co2e_factor,
        }),
```

When the next dataset ("2024-2025") is published we will have to spend some time addressing things like:
- download the correct csv based on the start_date and end_date period
- potentially deal with new/different column structure and formats
- correlate the locations of any new / changed facilities

It didn't seem worthwhile to deal with any of that right now, when we have enough data to actually proceed with processing 2023-24.

### DIFF
--- a/src/openmethane_prior/data_sources/safeguard/data.py
+++ b/src/openmethane_prior/data_sources/safeguard/data.py
@@ -55,7 +55,7 @@ safeguard_mechanism_csv_columns = [
 ]
 
 
-def parse_csv_numeric(csv_value: str) -> float:
+def parse_csv_numeric(csv_value: str) -> float | None:
     """Convert messy input values like " 124,138 " to float. Values of "-" are
     interpreted as None."""
     raw = csv_value.strip()


prayner:
src/openmethane_prior/data_sources/safeguard/__init__.py:
should this contain a license?
src/openmethane_prior/data_sources/safeguard/facility.py:
61: document the regexp?
122: there's nothing specific about a family or sector here is there
since it is looking in a list
tests/unit/test_data/test_anzsic.py
add license?
21: excellent choice of example
tests/unit/test_data/test_safeguard.py
add license?

aethr:
Thanks for the review @prayner.

> should this contain a license?

It's my understanding that you don't need to include a license header in every file, you only need to include one LICENSE with the distribution. Including the license header in files is useful in case someone is viewing a file in isolation who might not see the LICENSE. My feeling is, for files with no real logic or functionality in them (like __init__.py files that just define package exports), we don't really need a 17-line license block for a 2-line file.

I've taken the same approach to unit test files since it doesn't seem likely anyone would copy them, but I could probably be convinced that a license header is prudent since they contain a lot more code.

> 61: document the regexp?

Will do.

> 122: there's nothing specific about a family or sector here is there since it is looking in a list

Probably true. I've either removed or significantly reworked this method in a subsequent PR where DataFrame has been adopted as the end-to-end data representation for these datasets.

> 21: excellent choice of example

:joy: :beers:

Issue #97, opened 2025-07-30
Coal mines


Ida-droid:
## The motivation
Currently, openmethane-prior uses the ANGA total emission for Australia's fugitive emissions and regrids it using the Climate Trace dataset. In more detail, they normalise the Climate trace emission for each particular facility to obtain the relative contributions and then scale these by Australia's national fugitive emission number. 
They are not differentiating between mining emissions or oil and gas emissions. In the end, they are using the term "fugitive emissions." 

## Issues: 
Misallocation: 
The first issue is that mines have a spatial extent or source locations that are not at the facility location. 
For example, open-cut areas of the surface mines tend to extend spatially, sometimes several kilometres. Active mining in those open-cut areas produces most of the mine's emissions. These open-cut areas can thereby extend multiple grid cells, even in the 10 km spatial resolution grid, therefore leading to a misallocation of the mine's emissions if ignored.
Climate Trace is a good dataset to start with when using the mine's location, but sometimes, it is not up to date and might include mines that have already closed or neglect mines that have just been opened. 

Wrong prior emission: 
Climate Trace is unsuitable for obtaining each facility's relative contributions because it overestimates the opencut mine emissions by using too generous emission factors. 
Better prior emission datasets need to be included. 

## The proposed solution
1. There must be a roadmap of how and when to manually or automatically check the datasets: 
Is the facility still open? 
Do you know if the mine pit is still active? 
Is there a new mine pit? 
Is there a new ventilation shaft? 

Determine whether a mine is closed or operating: 
Mine locations, their owner, and parent ownership can be looked up in the following databases: 
https://globalenergymonitor.org/projects/global-coal-mine-tracker/tracker-map/
and https://portal.ga.gov.au/persona/minesatlas
These datasets are accurate, but sometimes don't account for a recent mine closure or opening. 
The reason is that sometimes, only news articles exist to determine whether a mine is closed. This poses another issue: you need to check if the mines are still operating manually. 

2. Update the emissions using Safeguard. 
Since this year (2025), many coal mines have been required to report their methane emissions in units of CO2eq to Safeguard.
The new Safeguard data covers the emissions for several facilities for the 2023-2024 period. 
 
Not all coal mines are safeguard facilities, but they must report to ANGA. 
Some of the mines that haven't reported to Safeguard are covered in the CSIRO dataset for the financial year 2020-2021: 
https://data.csiro.au/collection/csiro%3A62981v1
[...]

PR #141, opened 2025-10-28
Separate "fugitive" into "coal" and "oil and gas" sectors

### LINKED ISSUE #140: Separate "fugitive" layer into "coal" and "oil and gas"
## The problem

Relates to #97.

The current "fugitive" sector layer combines a coal mine dataset and an oil and gas facilities dataset into a single "fugitive facilities" list. It uses this list to spatialise the total emissions for the UNFCCC "1.B Fugitive Emissions From Fuels" sector.

However, the UNFCCC "1.B" category already breaks down neatly into:
- "1.B.1 Solid Fuels" (which almost entirely comprises of coal extraction and transformation)
- "1.B.2 Oil and Natural Gas"

This breakdown is also neatly reflected in the "060 Coal mines" and "070 Oil and natural gas" ANZSIC sectors.

By splitting this sector implementation, we can instead:
- spatialise the "1.B.1" total emissions by the coal mine dataset
- spatialise the "1.B.2" total emissions by the oil and gas facilities dataset

This may yield immediate improvements in our estimates by spatialising more discrete inventories. It will also make future improvements, including SGM integration, much simpler by working off a single ANZSIC code.

## Definition of "done"

The "fugitive" sector implementation has been removed, replaced by "coal" and "oil_and_gas" sectors which largely lift the existing implementation.

aethr:
## Description

Resolves #140.

This separates the "fugitive" sector, which spatialised UNFCCC sector "1.B" emissions using coal and oil & gas facility estimates, into two new sectors "coal" and "oil_gas":
- "coal" spatialises "1.B.1" emissions using coal facility estimates
- "oil_gas" spatialises "1.B.2" emissions using oil & gas facility estimates

This does not attempt to address any of the issues in #97 , but will hopefully make that simpler in the future.

## Checklist

- [x] Tests added

### Expected changes to results

Changes in outputs are expected due to splitting the emissions from the old "fugitive" sector into two new sectors. Annual total from the inventory for the test years:
 - fugitive: 1164824228.295256
 - coal: 888419995.6318997
 - oil_gas: 276404232.6633563

As expected, the `coal` and `oil_gas` sector totals equal the previous `fugitive` total. This means we should expect no change to the `ch4_total` values in the results over the full domain. However, we may see changes to `ch4_total` in subdomains due to the scaling of coal vs oil and gas sites based on the national inventory split, since coal appears 4x larger.

Looking at the respective sums in the coal and oil and gas input files gives us:
 - coal facilities sum (2022): 2638841.88
 - oil and gas facilities sum (2022): 556068.82

If we compare the proportions of coal vs gas in the inputs vs the inventory:
- inputs: 0.83 coal vs 0.17 oil_gas
- inventory: 0.76 coal vs 0.24 oil_gas

I believe this means we will see a slight reduction in emissions in the prior for coal sites, and a slight increase in oil and gas sites.

### Values in `au-test` domain

Recently we went to some amount of effort to identify a region for the au-test domain which contained values for every sector in the prior. Unfortunately, after splitting the fugitive layer we now have an "oil_gas" layer which has no facilities inside au-test. This is disappointing, but probablynot worth the effort revisiting at this stage.

### DIFF
--- a/changelog/141.improvement.md
+++ b/changelog/141.improvement.md
@@ -1,0 +1,1 @@
+Split "fugitive" sector into "coal" and "oil and gas" sectors


aethr:
> Lots of the numerical tests rely on exact equality. I've seen that break when one changes architecture e.g. big-little endian. You could consider np.allclose

Definitely valid, I'll make sure to keep this in mind when writing new tests. For the existing ones, I guess let's fix them up when it becomes a problem.

PR #144, opened 2025-11-10
Add Safeguard Mechanism facilities to coal sector

### LINKED ISSUE #94: Use emissions reported under the Safeguard Mechanism (SGA)

aethr:
## Description

Resolves #94.

This PR adds a Safeguard facility locations data source, which acts as a pivot table between facilities in the Safeguard Mechanism Baselines and Emissions Table and other reference data sources which contain locations for the Safeguard facilities. This file has been created initially with a small subset of SGM coal mines which can be expanded later.

In the coal layer, first Safeguard facility emissions are spatialised by:
- filtering out Safeguard facilities which do not match the ANZSIC code(s)
- filtering out Safeguard facilities when the emissions were reported outside the prior period
- finding the lat/lon for remaining Safeguard facilities in the existing "coal-facilities" data source
- allocating Safeguard emissions onto the grid using the lat/lon from "coal-facilities"

Emissions from any eligible facility, even facilities outside the domain of interest, are then substracted from the sector inventory total to yield an "unallocated" sum of emissions. This remaining sum is allocated to the remaining facilities in "coal-facilities", scaled by their relative contribution. In periods outside the Safeguard reporting period, this yields behaviour identical to the previous implementation.

## Checklist

- [x] Tests added

### DIFF
--- a/src/openmethane_prior/sectors/coal/sector.py
+++ b/src/openmethane_prior/sectors/coal/sector.py
@@ -17,7 +17,6 @@
 #
 
 import numpy as np
-import pandas as pd
 import xarray as xr
 
 from openmethane_prior.data_sources.safeguard import (
@@ -25,7 +24,6 @@ from openmethane_prior.data_sources.safeguard import (
     safeguard_mechanism_data_source,
 )
 from openmethane_prior.lib import (
-    DataSource,
     logger,
     PriorSectorConfig,
     kg_to_period_cell_flux,


aethr:
I've spent some time correlating the Safeguard Mechanism coal sector facilities with the existing `coal-mining_emissions-sources.csv` (Climate TRACE) dataset. I've been able to correlate 63 of the 70 SGM facilities with locations in our current dataset.

Some analysis of how we distribute 2023 emissions:

```
inventory emissions:  888,420 tCH4
safeguard emissions:  637,184 tCH4 (63 facilities)
remaining inventory:  251,236 tCH4 (28%)
CT total:     2,638,842 tCH4 (93 facilities)
CT remaining: 574,862 tCH4 (22%) (28 facilities)
```

Assuming our inventory figures for 2023 and the SGM figures align (which we hope they should!), the SGM facilities with locations account for 72% of the entire sector by raw CH4, leaving 28% unallocated. The coal mines in our Climate Trace dataset which are not correlated with an SGM facility make up 22% of the total from that dataset. These ratios seem close enough to indicate we're on the right track.

From an absolute numbers perspective, the facilities being allocated from the Climate TRACE dataset total 575Mt, versus the remaining inventory of 251Mt, so these locations will have their emissions reduced overall, but by a factor of roughly 2 which is consistent with our expectations that Climate TRACE overestimates these facilities. Since we were already scaling these numbers down based on the inventory total, I don't expect this will cause a major difference in the result.

### Missing facilities

3 of the 7 SGM facilities which I could not locate in our existing dataset were opened on or after the 2022 cutoff in our existing dataset:
- Centurion Coal Mine: redeveloped in 2022
- Ironbark No. 1: opened in 2023
- Olive Downs Complex: opened in 2024

The remaining 4 simply seem to be missing from the data:
- Goonyella Broadmeadow Mine: Goonyella Riverside is in our dataset, but not Broadmeadow
- Integra Underground Mine: appears on CT map but missing from our dataset
- Norwich Park Mine: re-opened in 2020 as Saraji South Mine, missing from dataset
- Russell Vale Colliery: re-opened in 2021, missing from dataset

aethr:
> can filter_locations be handled more elegantly with the kwargs mechanism and a loop over keywords rather than individual testing?

That is definitely one way of doing it, but it's not much of an abstraction. If the data structure were to change, all the code that uses it would also have to be updated and I don't think Python's type system would be that much help. I'm not against the idea if you prefer it though, after all there is plenty of other code where the data structure is "baked in".

> also, does filter_locations change locations_df or just return a restricted view?

I actually don't know the answer, I assume Python is pass-by-reference so re-assigning the variable in the method just creates a new reference. But re-assigning a value to an argument is poor practice anyway, so I should fix that.

PR #152, opened 2025-11-27
Add Safeguard Mechanism facilities to oil & gas sector 

aethr:
## Description

Similar to #144, this adds a proof of concept for allocating Safeguard Mechanism facility emissions for facilities in the "Oil and gas extraction (070)" sector to the `oil_gas` prior sector.

This is more complicated than the coal implementation due to a number of factors:
- An oil and gas "facility" in the SGM sometimes refers to multiple discrete locations
- Some of these locations are large oil or gas fields which can span multiple grid cells
- There is no single data source with locations for all Safeguard facilities

Currently, this implements a single new data source which gives the shapes of oil and gas fields in Victoria, Australia, and allocates the SGM facility emissions to them continuously across the entire area.

### Allocation of facility emissions

Even if it was clear exactly which fields and facilities were counted under a single Safeguard facility, the single annual emission figure must be allocated among them, which is a complex problem. How much do fields emit vs facilities? How much is lost in transport? Do pipelines leak at the start and end, or across their length? Answers to these questions may differ from facility to facility.

## Checklist

- [ ] Tests added


aethr:
Superseded by #170 .

PR #161, opened 2026-03-20
Replace Climate TRACE oil and gas locations with petroleum wells datasets

aethr:
## Description

As part of incorporating Safeguard Mechanism methane estimates into the prior, we must first build up a list of locations which correspond to petroleum titles/leases where emissions have been reported.

Most Australian states publish a public dataset of boreholes/drillholes which include bores drilled for petroleum production. By locating petroleum bores that occur within active petroleum mining title/lease areas, we can build a dataset of likely emission points within a given time period. While this certainly contains many false positives due to lack of published data when wells are capped/depleted, it is already a much more comprehensive dataset than our previous implementation.

One clear disadvantage of this approach is in allocating an amount of emissions to each point. The CT dataset includes an emission estimate for each location, which we can scale to the sector total from the ANGA inventory. The currently approach doesn't have any proxy for emission quantity, meaning emissions must be distributed naively across the board, which is bound to be inaccurate. However, this should be rectified when we integrate SGM emissions which will cover the majority of emissions in the sector.

This PR also does not include processing facilities or pipelines, which will be important to include for an accurate picture of emissions.

This PR supersedes #152 , but does not currently attempt to implement any integration of Safeguard Mechanism emissions, which will come in a later PR.

## Checklist

- [x] Tests added

## Notes

I ran an analysis comparing the emission sources derived in this PR with the previous Climate TRACE dataset. The CT dataset only contains 19 locations which roughly correspond to entire oil and gas fields. The new emissions sources dataset in this PR includes 9,946 geographic points corresponding to bores and wellheads within active petroleum titles.

Of the 19 existing locations, most have an equivalent point in the new dataset less than 1,000m away. The table below shows the distance from each point in the CT dataset to the closest point in the new approach.

```
                                        ct_source_name  ...      distance (m)
26                                    Brunello-Julemar  ...   2565.933387
56                    Codie Complex - coal bed methane  ...    615.937972
81                         Fairview - coal bed methane  ...    304.942613
146                                          Gippsland  ...  10818.745818
154                                     Greater Gorgon  ...   1505.872538
185                                            Ichthys  ...    306.965484
234                                             Janzlo  ...   5300.184748
281                                           Kingfish  ...    128.765761
299                                            Laverda  ...    757.421933
341                                        NWS Project  ...    355.384971
381                                       North Rankin  ...   2784.487544
396                                             Norton  ...    800.603382
441                                            Perseus  ...   3595.978279
481                                              Pluto  ...   1586.047955
499                                            Prelude  ...  11366.165974
514  Reedy Creek, Clifford East, Combabula, Lucky G...  ...    118.506366
515                           Surat - coal bed methane  ...     29.885828
576                   Talinga-Orana - coal bed methane  ...    712.293571
599                                         Wheatstone  ...   4637.132637

[19 rows x 4 columns]
```
The locations furthest away from an emission source in the new dataset are the Gippsland and Prelude, both representing offshore fields. After plotting the old locations on a map alongside points from the NOPTA (offshore) wells dataset, it appears that the CT points are possibly a "middle point" between many wells / fields in the area. However once the dataset is reduced to only "Development" wells, the CT points aren't necessarily that close to producing clusters.

Edit: updated analysis after adding NT well data.


prayner:
notebooks/README.md: I presume it's also possible to run the notebooks
outside jupyter with uv run python ...? If so probably worth adding
for us jupyter-phobes

.../sectors/oil_gas/data/nopta_titles.py: There are hardcoded urls in
here. Perhaps leave a note for future maintainers that if this
breaks because of changed locations then you should look here

.../sectors/oil_gas/data/nsw_drillholes.py: same as above

.../sectors/oil_gas/data/nsw_titles.py: Is that NSW dataserver url now
being used several times? Centralise?

.../sectors/oil_gas/data/qld_boreholes.py: url again

.../sectors/oil_gas/data/qld_leases.py and again

.../sectors/oil_gas/data/sa_wells.py: this is how I expected the urls
to be recorded

.../sectors/oil_gas/data/wa_titles.py: hardcoded url

.../sectors/oil_gas/data/wa_wells.py: hardcoded url

.../oil_gas/emission_sources/emission_source.py: The function name
normalise_emission_source_df. "normalise" to me means reset value so
that e.g. the sum is 1 or similar. This looks like it's just
reprojecting locations. If that's right I'd suggest capturing that in
the name.

.../oil_gas/emission_sources/nsw_sources.py: this holds for all these
regional files. Is it clearer to give the method a similar name and
just differentiate by the module? This might make it clearer they're
all doing versions of the same thing

aethr:
> notebooks/README.md: I presume it's also possible to run the notebooks
> outside jupyter with uv run python ...? If so probably worth adding
> for us jupyter-phobes

I've tried to carry the jupyter torch (which was added by the Climate Resource folks) but I'm honestly not sure I'm getting much use out of it. I've never tried to run these scripts via cli because fundamentally they're being used to explore data visually with maps and charts. I think a better alternative to jupyter exists (marimo) which stores notebooks in .py by default, so I might experiment a bit here and see if I can make these more valuable. If not, I'm inclined to delete the whole folder.

> .../sectors/oil_gas/data/nopta_titles.py: There are hardcoded urls in
> here. Perhaps leave a note for future maintainers that if this
> breaks because of changed locations then you should look here

My convention thus far has been to include a comment above every DataSource that includes a link to the origin of the data. The nopta_titles file includes this:

```
# Locations of all offshore petroleum and gas titles (licenses) administered
# by the National Offshore Petroleum Titles Administrator (NOPTA), via the
# National Electronic Approvals Tracking System (NEATS).
# Source: https://www.nopta.gov.au/maps-and-public-data/neats-info.html
```

For most DataSources which include a `url` param directly, this is as close as possible. Is your feedback that you'd like this to be closer to the hard-coded urls that are found in custom `fetch` methods?

> .../sectors/oil_gas/data/nsw_titles.py: Is that NSW dataserver url now
> being used several times? Centralise?

I don't feel strongly about doing this or not doing it. I would probably suggest that if we want to centralise the URL, we might move both of these DataSources and their supporting logic into a single file?

> .../oil_gas/emission_sources/emission_source.py: The function name
> normalise_emission_source_df. "normalise" to me means reset value so
> that e.g. the sum is 1 or similar. This looks like it's just
> reprojecting locations. If that's right I'd suggest capturing that in
> the name.

When talking about data structures I would expect "normalise" to mean "make them all the same". The intent here was a function which:
- ensures all emission sources include a minimum set of fields
- ensures all geometries use the same projection

This would allow us to combine multiple sets of emission sources and know we're dealing with a homogeneous structure. Do you have a suggestion for a word which would communicate that more clearly?

> .../oil_gas/emission_sources/nsw_sources.py: this holds for all these
> regional files. Is it clearer to give the method a similar name and
> just differentiate by the module? This might make it clearer they're
> all doing versions of the same thing

To clarify, are you suggesting that we rename `nsw_emission_sources` et al. to just `emission_sources` so that all regional implementations just implement a  `emission_sources` function? Certainly the intent in naming them all "*_emission_sources" was to make it clear that they're all doing versions of the same thing. If we gave them all the exact same name, what would the `all_sources` implementation look like? Something like:

```python
import .offshore_sources

    offshore_df = offshore_sources.emission_sources(
        start_date=start_date,
        end_date=end_date,
        offshore_wells_da=offshore_wells_da,
        offshore_titles_da=offshore_titles_da,
    )
```

prayner:
Replying to several comments at once here, let me know if there's a
more github'ish way of doing this.

> .../sectors/oil_gas/data/nopta_titles.py: There are hardcoded urls in
> here. Perhaps leave a note for future maintainers that if this
> breaks because of changed locations then you should look here
>
>My convention thus far has been to include a comment above every DataSource that includes a
>link to the origin of the data.
>For most DataSources which include a url param directly, this is as close as possible. Is
>your feedback that you'd like this to be closer to the hard-coded urls that are found in
>custom fetch methods?
I preferred the way you handled it in:
src/openmethane_prior/sectors/oil_gas/data/nt_titles.py
where the url was added to the DataSource instance. That's a form used
a lot through the code now. If that's not sensible then I'd suggest
noting the presence of these urls in a developers-notes.md or similar file.
>I don't feel strongly about doing this or not doing it. I would probably suggest that if we
>want to centralise the URL, we might move both of these DataSources and their supporting
>logic into a single file?
I think that's a good solution then define the variable at the module level.
>When talking about data structures I would expect "normalise" to mean "make them all the
>same". The intent here was a function which:
>
>* ensures all emission sources include a minimum set of fields
>* ensures all geometries use the same projection
>
>This would allow us to combine multiple sets of emission sources and know we're dealing
>with a homogeneous structure. Do you have a suggestion for a word which would communicate
>that more clearly?
My concern is probably only one for mathematicians. When I map
disparate sources on to a common data structure I tend to call the
functions Canonicalise but I realise that has its own freight in the
IT world. I'd leave it. 
>
> .../oil_gas/emission_sources/nsw_sources.py: this holds for all these
> regional files. Is it clearer to give the method a similar name and
> just differentiate by the module? This might make it clearer they're
> all doing versions of the same thing
>
>To clarify, are you suggesting that we rename nsw_emission_sources et al. to just
>emission_sources so that all regional implementations just implement a emission_sources
>function? Certainly the intent in naming them all "*_emission_sources" was to make it clear
>that they're all doing versions of the same thing. If we gave them all the exact same name,
>what would the all_sources implementation look like? Something like:
I'd import the relevant modules, put them in a list and iterate over
them calling their .emission_sources method. Or define a class and
instantiate it for each jurisdiction. This looks to me like a
good analogy with how we do process_emissions.

aethr:
> I preferred the way you handled it in:
> src/openmethane_prior/sectors/oil_gas/data/nt_titles.py
> where the url was added to the DataSource instance.

The `url` param for DataSource was originally an alternative for the `fetch` method, but since most of the fetch implementations retrieve a single URL anyway, I think it does form a nice sort of self-documenting pattern. I'll apply this across the board where appropriate.

>> move both of these DataSources and their supporting logic into a single file?

> I think that's a good solution then define the variable at the module level.

That being the case, I would probably combine each of the regional sets of wells/titles into a single file for each region and apply the same pattern.

> I tend to call the functions Canonicalise but I realise that has its own freight in the
> IT world. I'd leave it.

Standardise?

> I'd import the relevant modules, put them in a list and iterate over
> them calling their .emission_sources method. Or define a class and
> instantiate it for each jurisdiction. This looks to me like a
> good analogy with how we do process_emissions.

I don't think I agree in this case. There are three differences between this and `process_emissions`:
1. each sector's `process_emissions` method is completely independent, whereas the code that aggregates all emission sources doesn't treat each output the same. NOPTA sources have to be de-duplicated w.r.t. WA sources because of significant overlap.
2. Users can opt in/out of different `process_emissions` implementations by specifying `--sectors` on the CLI, and the looping mechanism / abstract API facilitate that. This makes the code harder to read, but serves a functional purpose.
3. Most importantly, each `process_emissions` takes identical arguments whereas the emission_sources methods all take similar, but actually quite different arguments.

I think the last reason is probably the strongest in favour of leaving the emission_source methods with different names. They're similar, but take different arguments and return subtly different outputs.

prayner:
agree with maintaining emission_sources names as they are

prayner (approved):

PR #162, opened 2026-03-25
Test using local docker image in GitHub Actions

aethr:
## Description

This PR makes two minor improvements to GitHub Actions:

1. Test using local docker image

The previous implementation built and pushed a docker image to the GitHub Container Registry, then specified it as the container image for the next job, causing it to be pulled down again. This round-trip is unnecessary as there is already a local copy of the built image.

Furthermore, using our app container as a GitHub actions container is problematic for other reasons. Actions must run as root, which prevents us from using a non-root user in our application. Also, using actions in the container to set up things like caches introduces subtle changes in the test conditions that we should avoid.

2. Add an input cache

The prior uses a number of remote data sources, and all of these must be fetched and processed in the course of running the tests. Using an input cache will make tests run faster in GHA, at the expense of some fetch/parse methods not being invoked on every test run.

## Checklist

- [ ] Tests added

### DIFF
--- a/.github/workflows/build_docker.yaml
+++ b/.github/workflows/build_docker.yaml
@@ -34,16 +34,19 @@ jobs:
     # Builds and pushes the image
     # Tags the image with the PR that it is linked to
     steps:
+      # Setup repo and docker utils
       - name: Checkout code
         uses: actions/checkout@v4
-      - name: Set up Docker Buildx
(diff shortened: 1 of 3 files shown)


aethr:
Merging without a review, as this is purely an infra change. Happy to take feedback to action later if any changes are necessary.

PR #167, opened 2026-04-09
Add oil and gas infrastructure and facilities

aethr:
## Description

Where #161 attempts to find oil and gas emission sources in oil and gas fields by locating all potential production wells, this PR attempts to add infrastructure and facility locations where emissions may also occur. These sites may include:
- oil refineries
- gas processing plants
- LNG terminals
- pipeline compressor units/stations
- on-site electricity generation

Infrastructure locations are taken from two new data sources:

1. TSI Oil and Gas Sites

This is a new data source created by researchers at The Superpower Institute, specifically focused on finding the locations of sites that comprise each Safeguard Mechanism "facility" in the 2023/24 SGM Baselines and Emissions report.

2. National Pollutant Inventory

The National Pollutant Inventory is a public resource provided by the Commonwealth of Australia, which includes locations of over 4,000 facilities across many ANZSIC sectors, where the facility has reported pollutant emissions of a range of substances covered in pollution legislation. While methane is not an explicitly tracked substance under the NPI, the locations of facilities in each ANZSIC sector can still be used to place emissions from industrial infrastructure and processes.

## Checklist

- [x] Tests added


aethr:
Tests are failing due to the NPI `facilities.csv` dataset download failing with a `403 Forbidden` response. This appears to be Cloudflare (fronting data.gov.au) blocking requests from basic HTTP agents such as Python `urllib` and `curl`.

It was working last week, so there may be a recent change in configuration. I plan to lodge a support request with Data.gov, but in the meantime I will probably upload a local copy of facilities.csv to our S3 bucket until data access at the primary source is restored.

prayner:
npi_data.py:27:
good to describe the return value in the doc string, I presume it's a
datetime for the first day of the FY?

npi/sources.py:31:
specify return type?


site_sources.py:96:
do you need to raise an error or is there a simple transformation that
can get you to the right units?

prayner (approved):

aethr:
> npi_data.py:27:
> npi/sources.py:31:
> specify return type?

Have improved the docstrings and type annotations for all the noted functions. In general I prefer to provide type annotations over docstrings, and I'll omit a docstring where I think the return type and the function naming are sufficient to describe the intent. I'll generally only use a docstring where I think the intent is ambiguous, but I'm aware the ambiguity is in the eye of the beholder. :)

> site_sources.py:96:
> do you need to raise an error or is there a simple transformation that
> can get you to the right units?

This relates to the `max_distance=250` on line 105, which is used to discover which NPI facilities are in close proximity to existing sites, and are therefor likely to be duplicates. The `max_distance` param must be in the unit of the CRS, but for us 250m has been determined to be a reasonable threshold.

I think you described an approach that would let us approximate "250m" in the supplied CRS, but I didn't see it as a high priority since all of our domains use meters. Raising an error was my way of saying "if this error happens, lets implement the 250m conversion".

I'll see if Claude can come up with something quickly, otherwise I'm inclined to leave this as a problem for the future.

aethr:
The solution was staring me in the face: reproject both into a CRS that uses metres just for the distance check, keeping the index but throwing away the reprojected coordinates.

Issue #165, opened 2026-04-08
Mapping the "27 Gas supply" ANZSIC sector


aethr:
## The problem

The oil and gas sector attempts to spatialise emissions covered under UNFCCC code "1.B.2", which includes, for both oil and natural gas:
- Exploration
- Production and Upgrading
- Transport
- Refining
- Distribution of Oil Products
- Other

In ANZSIC, these activities are spread across multiple sector codes:
- [07 Oil and Gas Extraction](https://www.abs.gov.au/statistics/classifications/australian-and-new-zealand-standard-industrial-classification-anzsic/2006-revision-2-0/detailed-classification/b/07/070/0700)
- [17 Petroleum and Coal Product Manufacturing](https://www.abs.gov.au/statistics/classifications/australian-and-new-zealand-standard-industrial-classification-anzsic/2006-revision-2-0/detailed-classification/c/17/170)
  - Although this includes "coal products", from a practical standpoint, very little coal product mfg occurs in Australia
- [1811 Industrial Gas Manufacturing](https://www.abs.gov.au/statistics/classifications/australian-and-new-zealand-standard-industrial-classification-anzsic/2006-revision-2-0/detailed-classification/c/18/181/1811)
  - This includes all industrial gases, but in practice the majority of facilities appear to be related to LNG/LPG processing/supply
- [27 Gas Supply](https://www.abs.gov.au/statistics/classifications/australian-and-new-zealand-standard-industrial-classification-anzsic/2006-revision-2-0/detailed-classification/d/27/270/2700)

Currently, locations falling under 07,  17, and 1811 sectors are mapped using a combination of well locations and production facilities found in the NPI, resulting in a long list of point source emissions.

Facilities classified as "27 Gas Supply" represent 27% of Safeguard facility emissions in this sector, but spatialising these emissions is much more difficult. If we can't allocate these spatially, those emissions will be allocated to the non-SGM point sources for the rest of the sector, inflating emissions for those locations across the board.

### 1. Pipelines

Geoscience Australia has published a geospatial dataset of both oil and gas pipelines across Australia. While the dataset includes 1446 gas pipelines, it is fairly limited in scope:
- appears to only include transmission lines, no gather lines or distribution lines
- includes the state the pipeline operates in, but no ownership details

Spatialising emissions along pipeline shapes would be better than nothing, but has many drawbacks:
- cannot effectively allocate SGM facility emissions without manually correlating pipeline entities with corporate ownership
[...]

PR #170, opened 2026-04-16
Add Safeguard Mechanism facilities to oil & gas sector 

aethr:
## Description

Supersedes #152.

Allocate methane emissions reported under the Safeguard Mechanism to facilities in the oil and gas sector (UNFCCC sector 1.B.2, ANZSIC sectors 07, 17 & 27).

This PR builds on #161 and #167, where we have built an entire database of possible emission sources in the oil and gas sector from petroleum wells, infrastructure and processing sites. Using the `safeguard-locations` data source, we've related each SGM facility with:
- locations of industrial sites and infrastructure that are likely to be classed as part of the "facility"
- petroleum production titles where oil and/or gas extraction is solely tied to the facility

Methane emission reported for each facility is then distributed evenly amongst all the wells and sites associated with that facility. This approach is incredibly naive, but will have to suffice until we can implement a better strategy.

Finally, the methane allocated to SGM facilities is removed from the ANGA sector total giving us a remainder of unallocated methane, which is spread evenly across all the remaining emission sources that were not associated with an SGM facility.

## Checklist

- [ ] Tests added


aethr:
This implementation exhibits a numerical issue that we're investigating. The sector implementation spatialises the emissions from UNFCCC category `1.B.2` (Fugitive emissions from fuels, Oil and Natural Gas) to facilities from ANZSIC sectors `07`, `17`, and `27`, and sites and wellheads from other data sources.

```
Nat. inventory - UNFCCC 1.B.2: 232.16 kt CH4
Safeguard facility total:      105.70 kt CH4
```

Safeguard facilities within these ANZSIC sectors only account for 45.5% of the `1.B.2` category national emissions, a lower proportion than expected in such a highly concentrated industry. At the same time, based on our current data and approach, 66.8% of our 11,124 potential emission sources are being associated with a Safeguard facility. This leaves the majority (54.5%) of the total emissions to be distributed to a small number (33.2%) of remaining locations. This would cause non-Safeguard locations to appear to have almost 2x the emissions of equivalent locations in the Safeguard facilities.

```
3694 / 11124 non-SGM sources (33.2%)
126.46 kt non-SGM emissions (54.5%)
155.27 kt unallocated emissions including ANZSIC 27 (66.9%)
```

Below are some contributing factors and possible mitigations for this discrepancy. 

## Inability to spatialise ANZSIC 27 Gas Supply facilities

As noted in #165, we currently have no strategy for locating Safeguard facilities in the ANZSIC 27 Gas supply sector, representing 28.81 kt of CH4 emissions (12.4% of the sector total). These emissions end up being counted as "unallocated" and distributed amongst the remaining sources, exacerbating the issue.

## Inclusion of capped and suspended wells

Our approach to identifying emission sources from wellheads suffers from its inability to distinguish between wells that are producing, suspended or capped. In theory producing wells are more likely to produce fugitive emissions, but borehole datasets only reflect the current status of each bore at the time of the most recent update. A bore listed as capped in 2026 may have been producing in 2023, but the data lacks this information.

This leads to our 11,124 potential emission sources being dominated by suspended or capped wells in active leases. As an example, these are the "status" values for the boreholes in QLD which we consider potential emission sources in FY 2023/24:

```
SUSPENDED                8138
PRODUCING                1677
PLUGGED AND ABANDONED     726
COMPLETED                  32
```

It's worth considering the [NGER legislation](https://www.legislation.gov.au/F2008L02309/latest/text) which provides the measurement framework behind both the national inventory and Safeguard figures we're attempting to spatialise. NGER only contains methods for estimating emissions from **producing** wellheads (see Part 3.3—Oil and natural gas—fugitive emissions). Our approach then is distributing reported emissions from a small number of producing wells to a large number of producing/suspended/capped wells, leading to much smaller per-well emissions. In fact if capped wells were still emitting methane this would be **on top** of the methane reported under NGER.

Lack of accurate well statuses was a known issue, with the rationale that wells in a field are close enough that most are likely to fall in similar/adjacent grid cells where the emissions will just be aggregated back together again. What we didn't anticipate was the over-representation of non-producing wells in Queensland, which end up absorbing much of the national emission due to sheer number of points.

Simply filtering `SUSPENDED` bores from the QLD dataset gives us a much more realistic picture of emissions sources after Safeguard emissions have been allocated:
```
2801 / 3962 non-SGM sources (70.7%)
126.46 kt non-SGM emissions (54.5%)
155.27 kt unallocated emissions including ANZSIC 27 (66.9%)
```

After allocating SGM emissions to SGM locations, we still have 70.7% of emission sources left to distribute the remaining 54.5% of emissions, leading to smaller emissions for locations outside the SGM. Although this looks like a good solution (and it could be a short term fix), this approach would not work well for historic periods. The further back in time you run the prior, the more inaccurate the current well statuses would be.

## Queensland over-representation

In the current implementation 8,732 of the 11,124 emission sources identified in the FY 2023-24 period are in Queensland. This over-representation is probably due to inclusion of suspended wells in a state where oil and gas have been developed for decades. This is especially problematic when we apply our simplistic method of distributing the total emissions evenly across all potential emission sources, effectively distributing 75% of all emissions to QLD.

Although ANGA withholds state-based inventories for the oil and gas sector based on "confidentiality", QLD is the one state where a sector inventory is actually available; according to ANGA QLD accounts for 85.99 kt or 37.0% of the national total of oil and gas emissions.

We could treat QLD as a special case and apply our existing methodology to sources in the state, subtracting QLD-based Safeguard emissions from state inventory total and distributing the rest to remaining emission sources in QLD. Then you continue the existing approach taking all emissions that haven't been allocated to SGM or QLD sources, and distributing to all remaining emission sources. 

Applying this approach gives us:

```
# Within QLD
1566 / 8768 non-SGM QLD sources (14.1%)
62.00 / 85.99 kt non-SGM QLD emissions (72.1%)

# National excluding QLD
2128 / 2356 non-SGM sources excluding QLD (90.3%)
93.27 / 146.17 kt non-SGM emissions excluding QLD (63.8%)
64.46 / 117.36 kt non-SGM emissions excluding QLD + ANZSIC 27 (54.9%)
```

You can see that the proportions in QLD are still pretty bad (72% of emissions going to 14% of the locations), but the remaining national distribution looks more sane (64% of emissions going to 90% of the locations).

## Naive distribution

The last clear issue which exacerbates this problem is our naive approach to distributing a single emission number among a large number of emission sources. Common sense suggests that if a central processing facility gathers oil or gas from 10s or 100s of wells, that emissions are likely to be much higher at the facility than at each individual wellhead. The NGER measurement framework bases most Method 1 approaches on volume passing through equipment, which would be higher at a facility which aggregates resources from multiple wells.

Like the well status issue, it was assumed that most processing facilities and compression stations would be co-located in fields where incorrectly attributed emissions would simply be re-aggregated again into a small number of adjacent grid cells. However, in practice we find that some fields have a much smaller number of wells than others, even when they produce similar volumes. Without using field production figures, we lose this information and distribute the highest emissions to the fields with the most wells.

If we allocated a smaller proportion of emissions to wells and a larger proportion to facilities, this would reduce the severity of the problem. Potentially we could review the NGER Method 1 calculations to try and extract emission factors for different types of infrastructure (oil vs gas, wells vs processing, etc), then apply these as weights by aligning the emission source `site_type` with equipment types present in NGER. This would be a fairly large job, however.

## Conclusion

I'll talk to the team about these findings and then make a plan to move forward. Watch this space.

aethr:
Had a good conversation with @prayner today, out of which have come some decisions:

- Spatialising ANZSIC 27 Gas Supply facilities
  - We agree that this should be tackled as a priority
  - Details on the approach will be in #165 
- Inclusion of capped and suspended wells
  - We agree that capped / suspended wells should NOT be allocated inventory emissions
  - Removing these improves the quality of our short term results
  - Emissions from abandoned / suspended wells would occur in addition to inventory, if we want to model this, it should be separate.
  - This does introduce a temporal issue running the prior on historic periods where well status is likely to be significantly different from current status in datasets
- Queensland over-representation
  - We agree that having the data to do QLD "separately" does help mitigate some of the other issues
  - Exploratory code here will be quick to "productionise", seems like low effort and risk but high reward
- Naive distribution
  - We agree **not** to try and tackle this problem at this stage
  - Our current approach is probably wrong, but its simple and requires no additional effort
  - We probably lack the data and expertise to do it really well
 
All of this is currently blocking re-starting production daily runs, so getting these pieces resolved is high priority.

aethr:
Today's update:

## Inclusion of capped and suspended wells

Contrary to the decision yesterday, removing `SUSPENDED` wells turned out not to be a good step forward. At least in the QLD boreholes dataset, the `SUSPENDED` status is not synonymous with "not producing". The QLD petroleum production dataset includes a column for "WELLS_ON_PROD" for each field/report, and excluding `SUSPENDED` bores from the dataset results in leases with far too few wells to match the number of producing wells for each lease.

To get our active wells closer to matching the production dataset, we end up needing to include: `PRODUCING`, `COMPLETED` and `SUSPENDED` statuses. We also needed to include wells with `result` value of `UNKNOWN` which actually ended up adding more wells into the emission sources.

This puts us back to our original statistical issue: in QLD, 66.43 kt out of the state total 85.99 kt is not coming from Safeguard facilities, but 85% of the wells in the boreholes dataset are on leases that are associated with Safeguard facilities. We don't understand the industry well enough to understand what types of facilities / activities are responsible for the remaining 66.43 kt of methane.

Some possible explanations:

1. Safeguard facilities don't include well emissions

Perhaps we've misunderstood or misinterpreted NGER facility definitions, and the listed facilities only refer to gas plants, processing facilities and refineries. This would move all 11,000 wellheads out of the SGM and we could spatialise the 66.43kt of methane across all of them equally.

2. Numerous downstream facilities emit large amounts of methane

Another explanation is that methane is emitted by downstream pipeline and facility operations, and that these facilities are both:
- owned / operated by non-Safeguard companies (or companies that represent shared interests between multiple upstream stakeholders)
- are numerous enough that their aggregated emissions end up being 4x the upstream emissions where oil/gas is extracted

Additionally, it's worth noting that there are no Safeguard facilities in QLD in the "27 Gas supply" or "502 Pipelines and other transport" ANZSIC categories (ignoring the QLD->SA pipeline which is mostly in SA). However, we can see from the NSW and VIC counterparts that emissions from these sub-sectors can be substantial. Unfortunately, there is currently no easy way for us to distribute emissions to pipelines / supply networks. Tackling this in #165 may partially address this problem.

aethr:
## Improving on our naive distribution

As above, our implementation breaks oil and gas production down into a list of "emission sources", which includes everything from wellheads to gas plants and oil refineries. The simplest approach to spatialising a single emission number is to divide it equally between all sources. However, conventional practice tells us that emissions from a single wellhead will be much smaller than emissions from a gas plant or oil refinery which processes output gathered from 10s or 100s of wellheads.

Unfortunately, improving on this naive approach is difficult without more data about activities and production variables at each location.

### Distribute using wellheads vs facilities

One clear distinction in our emission sources dataset is between wellheads and "everything else". Every eligible point in the bore/drillhole/well datasets can be considered a wellhead. Our other datasets (oil and gas sites and NPI facilities) don't include any wellheads, giving us a single clear delineation between extraction and processing locations.

Unfortunately there's not much data to classify/delineate between different types of facilities; some are small (compressors, etc) while some are large (refineries). However, even if we can only distinguish between wellheads and "facilities", we may be able to adopt a simple change to our distribution approach that yields a large improvement on distributing emissions.

**Assumption 1: Emissions happen at the point of extraction and at the point of processing** 

If we consider that all unprocessed material extracted at a wellhead must be processed at at least one facility, we can surmise that for each unit of "saleable" resource, emissions can occur at a minimum of two locations: the wellhead where extraction occurs, and at least one facility where it is processed. Because we don't have detailed information about the exact web of resource flow from wellheads through facilities to their eventual downstream destination, we might take the naive assumption that each unit of resource emits at one extraction point and one processing point.

**Assumption 2: Emissions scale with quantity of resource flowing through equipment**

Most Method 1 calculations in the [NGER Measurement Determination](https://www.legislation.gov.au/F2008L02309/latest/text) are based on the "quantity of fuel type" or "quantity of natural gas" which "flows through" the equipment/tank/facility. For this reason, if one facility processes oil or gas from 10 wells, its emissions would be calculated from the total output of all 10 wells.

**Calculation: Divide emissions into extraction and processing**

Effectively, the approach would look like:

```
total emissions = extracting emissions + processing emissions
```

`Extracting emissions` would be allocated to wellheads, `processing emissions` would be allocated to facilities. Even if we don't apply any proportional scaling factors to either figure, and simply divide total emissions evenly between the two categories, this will allocate a larger share of the emissions to processing facilities than wellheads.

For example, with our current data, QLD and national splits are:
```
QLD wells: 11334
QLD other: 145
National (excl. QLD) wells: 2148
National (excl. QLD) other: 603
```

The same approach could be applied both within individual Safeguard facilities which include wells and facilities, and to the remaining unallocated sources and emissions.

While this approach is hard to verify, its simplicity makes it easy to adopt and explain. It can certainly be improved, but at face value it seems to be an improvement over our naive approach.

### Distribute using NGERS emission factors

The [NGERS Measurement Determinations](https://www.legislation.gov.au/F2008L02309/latest/text) provide the calculations which companies must use to estimate their own emissions. Most Method 1 calculations scale a production amount against an emission factor to yield an emission amount.

While we don't have access to the production volumes at each facility/piece of equipment, another approach might be to use the legislated emission factors to develop an approximation that would allow us to scale or weight certain types of facility or infrastructure to take a relative proportion of emissions.

This approach probably isn't feasible for a number of reasons:
1. Different production variables can be used even within a single facility, ie unprocessed natural gas for some activities, processed natural gas for others. The resource extracted at each facility will have different proportions of saleable resource per unit extracted, making these impossible to generalise.
2. "Saleable" products may be processed and stored many times between extraction and final output, and the amount of processing is different from facility to facility. A final emission number can't realistically be broken down in a way that represents this accurately.
3. The approach is impossible to verify. If we take an approach with this much complexity, we need to be able to attest to its accuracy, and we have no way of doing that.

### Proposal

I propose we implement the wellhead vs facility allocation strategy with the idea that this could be improved at a later date. It should yield a substantial short-term improvement to how unallocated QLD emissions are distributed (which is currently a serious issue), and it's simple enough to explain that we can justify its adoption.

prayner:
General comment:
As far as I can tell this successfully implements the algorithm as I
currently understand it.

You've used nan often to mean "undefined". It at
least used to be true that not all machines even had nans and when
they didn't the result of assigning nan was undefined ... but often zero.
I doubt it's worth building extra masks for that but is it possible
the underlying arrays are masked ndarrays already?

emission_source.py:83: why are those types quoted?

oil_gas/sector.py:
170: have we played this game often enough to suggest an abstraction?

prayner (approved):

aethr:
> You've used nan often to mean "undefined". It at
> least used to be true that not all machines even had nans and when
> they didn't the result of assigning nan was undefined ... but often zero.
> I doubt it's worth building extra masks for that but is it possible
> the underlying arrays are masked ndarrays already?

Pandas recommends using these placeholder values to represent "[missing data](https://pandas.pydata.org/docs/user_guide/missing_data.html)". There are special placeholders for missing dates/times (`np.NaT`) and non-numpy dtypes (`pd.NA`), but I think since we're using numpy floats, `np.nan` is the correct placeholder.

We then explicitly use `np.isnan()` to construct masks (or `np.isnat()` or `pd.isna()` in some places depending on the dtype), so I believe this usage is safe and differentiates these rows from rows with a value of `0.0`. I did various checks to ensure it was returning the right rows, and I believe there are some unit tests somewhere that verify this.

> emission_source.py:83: why are those types quoted?

Good question, although Python's type system recognises `pandas.Series[bool]` as a `Series` of `bool` values, `pandas.Series` isn't actually implemented as a "generic type" (allowing the [contained_type] syntax). This is a valid Python type annotation that's visible to the type checker but "hidden" from the Python interpreter at runtime.

There's a package called "pandas-stubs" which I've installed in another branch which gives better types for Pandas classes which I'll probably start using soon.

> oil_gas/sector.py:
> 170: have we played this game often enough to suggest an abstraction?

I assume you're referring to the "iterate through a list of sources and allocate them to the grid" pattern? Interestingly, I replaced this exact loop in the new livestock layer (#183) with a single call to `np.add.at` (which you showed me at one point) and performance improved significantly.

However, in this particular case, I anticipated the need for a separate allocation strategy based on the geometry type, so we could support allocating to LineString or Polygon geometries. Although this hasn't been used yet, I might try to use it for the pipelines implementation. If it's not needed I'll try to condense this down to the `np.add.at` pattern.

PR #177, opened 2026-04-27
Fix build version check to pass required `TAG_VERSION` env var

aethr:
## Description

The container version check uses a TAG_VERSION env var which is no longer available now that the version check is run using `docker run`. This change passes the necessary env var to `docker run` via the `-e` flag.

This wasn't caught in any PR builds because this step only runs on version release builds.

## Checklist

- [ ] Tests added


aethr:
Merging without review, as this is just a small fix to the build configuration.

PR #184, opened 2026-05-08
Update wetlands sector to use updated SatWet-GIEMS dataset

### LINKED ISSUE #138: improvement of wetland fluxes
## The motivation
There's a fair bit of evidence that the wetland fluxes for Australia
are poor. Most datasets can't capture the unusually large seasonal
cycle and tend to overestimate fluxes. There is a new product driven
by improved satellite data which may address this.

## The proposed solution
Replace the current wetland flux with the improved product

## Additional context

The body of an email from Juliette Bernard:

Dear Peter,

I case this is useful to you, here is a link to a new version of 
SatWetCH4 estimates :

<private link redacted>

- a file called 'SatWetCH4_GIEMS-MC.nc', containing monthly methane flux 
estimates from the SatWetCH4 run using the GIEMS-MC wetland extent database.

- a short document explaining the changes made compared to the 
publication https://doi.org/10.5194/gmd-18-863-2025.

The current estimates range from 1992 to 2020, 0.25°x0.25°. An extension 
of GIEMS-MC until 2022 should be available within the next few months, 
which will also extend the SatWetCH4 estimates. These results are 
preliminary, we are planning to make the outputs available online within 
the next few months.

aethr:
## Description

Resolves #138.

Updates the wetlands sector implementation to use the updated wetland emissions dataset from [Satellite-based modeling of wetland methane emissions on a global scale (SatWetCH4 1.0)](https://doi.org/10.5194/gmd-18-863-2025).

This is based on #186, which generalises the conservative regridding method and adds major performance improvements.

## Checklist

- [ ] Tests added

## Questions

1. Shall we continue to construct monthly climatology by averaging each month over all available years?

**Answer:** if the period of interest is within the period of available data, the monthly SatWet data for the period should be used directly, since SatWet is constrained by actual data. If the period of interest falls outside the available data, using climatology is still recommended to smooth any annual anomolies.

2. Shall we still apply land masking?

The updated dataset appears to use its own land mask, including large inland lakes. I can confirm that applying our own land mask does reduce the total amount of CH4 in the layer, so we must be masking some grid cells which aren't masked in the source dataset.

**Answer:** continue to apply land mask, since this is aligned with our existing practices.

## Notes

Integration tests will fail because the dataset won't be uploaded to our public data store until we receive confirmation that the research has been published.


aethr:
Initial aim is to make minimal changes to use the new dataset as a drop-in replacement for `DLEM_totflux_CRU_diagnostic.nc`. Updates to `make_wetland_climatology` enable passing either DLEM_totflux or SatWet-GIEMS as a DataAsset, and using the appropriate variable names for each.

In addition, since `DLEM_totflux` is already in kg/m²/s, support was added for converting from gCH₄/m²/month if SatWet-GIEMS is used:

```python
    for month in range(12):
        climatology[month, ...] = flux[month::12, ...].mean(
            axis=0
        )  # average over time axis with stride 12

        # SatWet is in gCH₄/m²/month, convert to kg/m²/s
        if ch4_var == "fch4_mean":
            _, days_in_month = calendar.monthrange(2021, month + 1)
            climatology[month] = climatology[month] / 1000 / days_in_month / SECONDS_PER_DAY
```

Assuming no other changes are necessary, the difference in results is significant. The total emissions from wetlands using SatWet-GIEMS is smaller by a factor of 10. The prior was run with the same configuration over the aust10km domain and results compared:

```
sum:
 DLEM:   3.1493654859935395e-06
 SatWet: 2.9395411841310503e-07
max:
 DLEM:   7.551483327539188e-10
 SatWet: 8.838244025533015e-10
mean:
 DLEM:   1.6132391588943446e-11
 SatWet: 1.5057582133649474e-12
standard deviation:
 DLEM:   4.234337683454582e-11
 SatWet: 1.5591161917798652e-11
```

prayner:
The different relationships between the max and the mean is
interesting, the max is not very different while the mean is. An
important variable among different wetland methane estimates has been
wetland extent. I wonder how much of the differences this would
explain. It can be probed with tests like (emis > threshold).sum()
where emis is from a given data set and threshold is, say 1e-14.
nonzero = (emis > threshold) and
emis[nonzero].mean() will also pick whether the differences in
emission intensity for emitting points or the sheer number of points
emitting (wetland extent) is the critical variable.

aethr:
```python
    threshold = 1e-14
    print(f"above {threshold}:")
    print(f" DLEM:   {(dlem_xr['ch4_sector_wetlands'] > threshold).sum().item()}")
    print(f" SatWet: {(satwet_xr['ch4_sector_wetlands'] > threshold).sum().item()}")
```
result:
```
above 1e-14:
 DLEM:   68193
 SatWet: 22888
```

Ran the same diagnostics on the non-zero masked result:
```
sum:
 DLEM:   3.21362431018517e-06
 SatWet: 2.939466248095166e-07
max:
 DLEM:   7.551483327539188e-10
 SatWet: 8.838244025533015e-10
mean:
 DLEM:   4.712542797919391e-11
 SatWet: 1.284282701894078e-11
standard deviation:
 DLEM:   6.022615183000567e-11
 SatWet: 4.390619917889261e-11
```

Probably worth nothing that I think the DLEM dataset has a number of large negative values, which may also affect its statistics post-masking.

aethr:
@prayner this is ready to review. The PR is still in Draft status, as I haven't uploaded the SatWet dataset to the public data store.

Have we had explicit permission from Juliette to publicly release the dataset she provided? Do we need to wait for her to publish?

prayner:
wetlands/sector.py: the code seems to calculate the climatology
unconditionally although you've tested earlier whether it's necessary.
Probably only minor performance hit?

prayner (approved):

aethr:
> wetlands/sector.py: the code seems to calculate the climatology
> unconditionally although you've tested earlier whether it's necessary.
> Probably only minor performance hit?

That's true, since each time step will be evaluated independently in the loop which assigns output values. I could simply move the step which calculates the climatology inside the previous date check. Then during the loop where values are assigned I can raise an error if the date is outside the input date range but climatology isn't available (although this should never happen).

It's a minor performance gain but it's so simple I think we may as well do it!

aethr:
Approved and ready to merge, now just waiting on approval to use the dataset.

aethr:
@prayner and I have decided to merge and release this change, and publish the dataset which was shared with us in our public data store. We're still engaged with the researchers who provided the dataset to secure explicit data licensing for Open Methane users, but believe that we have implicit approval to share the data publicly.

We'll update our data sources documentation with more details on the publication and data licensing as more info becomes available.

Typically we would wait for explicit licensing before taking these steps, but the difference in results between this dataset and the previous wetlands dataset is large (a factor of 10) and we would like to include this in our forthcoming data processing.

aethr:
There's a discrepancy between early results using this dataset and the current results using the merged code in `main`. Early results were calculated in a temporary branch which included the SatWet dataset as well as some other changes, but did not include zeroing NaN/masked cells (#198). Because the SatWet dataset doesn't include any negative values in Australia, we'd expect early results without zeroing to be larger than current results, but the opposite appears to be the case.

```
2024-01-01 - 2024-01-31:
- early['ch4_sector_wetlands'].sum()   = 1.27228298e-05
- current['ch4_sector_wetlands'].sum() = 1.39497676e-05 (+9.64%)

2024-06-01 - 2024-06-30:
- early['ch4_sector_wetlands'].sum()   = 8.41960208e-06
- current['ch4_sector_wetlands'].sum() = 9.54387812e-06 (+13.35%)
```

It turns out this was caused by how NaN cells are treated during conservative regridding. Previously, a cell in the target grid which overlapped multiple cells in the source grid would be set to NaN if any of the intersecting source cells were NaN, losing the values from the other intersecting cells. This was fixed in #198 by converting source NaN values to zero prior to regridding, allowing a correctly conserved value to be calculated for each target cell.

Now that values in the other intersecting cells aren't being lost, the total sum of the layer is higher.

PR #185, opened 2026-05-11
Balance oil and gas emissions between wells and facilities

aethr:
## Description

This PR implements an emission allocation strategy outlined in [this comment](https://github.com/openmethane/openmethane-prior/pull/170#issuecomment-4402698664).

Previously, all emissions being allocated to a set of emission sources were distributed evenly among all sources. In the case of a facility with a mix of wellheads and facilities, this resulted in equal emissions at a single wellhead, and at a central processing facility handling materials from 10s or 100s of other wellheads. Taking NGER Method 1 calculations for oil and gas activites as a guide, emissions are typically scaled based on the amount of raw material passing through each piece of equipment. A central processing facility handling material from multiple wells should therefore have higher emissions based on the larger aggregated amount of material being processed.

This PR makes an assumption that for each unit of material extracted at a well, equal emissions will happen at the wellhead and at a single processing facility. When there are more wells than facilities, the emissions at each facility are scaled based on the number of wells which might "supply" that facility. Since we don't have a data source to tell us which wells supply which facility and so forth, we use a simple weighted average.

This is still a very simplistic approach, but it ensures that facilities emissions are scaled to reflect the greater volume of resources that pass through their equipment.

## Checklist

- [x] Tests added


aethr:
This could probably be simplified and made more performant by adding an `is_wellhead` boolean column to the emission source DataFrame, instead of filtering based on `site_type` string values that start with "drillhole". If we decide to adopt this approach I'll probably make the extra effort to make that change.

aethr:
This PR has been superseded by #191 which builds on this approach but also includes balancing of emissions from pipelines.

PR #186, opened 2026-05-13
Refactor and generalise conservative regridding in fire, termite and wetlands

### LINKED ISSUE #72: Evaluate xESMF as a replacement for custom regridding methods
## The motivation

Regridding from source datasets in the wetlands, termite, fire and other layers is currently done using bespoke regridding methods, some of which are very similar from sector to sector.

The ESMF (and xESMF Python module) appear to be a standard tool for regridding in the Earth sciences, and they support several key features that we care about:
- offers a conservative regridding algorithm
- separate steps for 1. creating the "regridder" (ie regridding weights) and 2. applying regridding to a dataset
  - this should allow us to create a regridder for each input domain to the solution domain and save it to disk for re-use in successive runs
 
Using an available open source solution for this task has several advantages:
- solution has better documentation
- maintenance burden is shared amongst open source contributors (we have less code to maintain/test)
- open source solution is more likely to be robust and performant with more eyes across its implementation
- new contributors to openmethane-prior may already be familiar with the tooling

## The proposed solution

To evaluate if xESMF is suitable for openmethane-prior, we should identify one or more existing layers, and implement a side-by-side regridding approach using the existing implementation and xESMF and compare the results. If the regridding output is within acceptable tolerances, we can conclude that xESMF is suitable for our purposes.

Converting each bespoke re-gridding method to use xESMF can follow later.

## Alternatives

Several other alternatives were considered, such as:
- xarray-regrid
  - This appears to be too rigid in its input requirements, and not widely adopted
- create an internal shared solution
  - This may be better than the current state of having multiple lightly-tested implementations of the same algorithm, but still requires us to write and maintain code for what is a very common usecase.

aethr:
## Description

The `fire`, `termite` and `wetlands` layers are all based on gridded geospatial data which must be interpolated onto the domain grid while conserving values. Each sector currently re-implements nearly identical functionality for achieving this.

This PR replaces the three conservative regridding implementations with a single shared helper, and improves performance and caching by utilising scipy sparse CSR array which is designed for this sort of grid weight data.

## Checklist

- [x] Tests added

### Branch base

This branch was originally stacked on the wetlands SatWet branch, as it grew out of some refactors made during that conversion. However, this change should be adopted whether or not we proceed immediately with SatWet integration. I've rebased this branch onto a mainline branch so it can be reviewed and merged in isolation. I'll rebase the SatWet branch onto this and it will end up looking much smaller.

### Performance

Note: these results were based on the SatWet version of this branch, but the results should be comparable with any data source.

A ran several test runs to gauge the impact of these changes on performance. The tests used `aust10km` domain over a period of 3 days, running only the wetlands layer, all inputs were from file cache. Tests were run on the SatWet integration branch (#184) before these changes were introduced, and on this branch (#186).

- Dates: 2023-07-01 to 2023-07-03 (uses climatology)
  - SatWet - no cached weights: 119.9s
  - SatWet - cached weights: 96.0s
  - This branch - no cached weights: 42.3s
  - This branch - cached weights: 18.9s
- Dates: 2022-07-01 to 2022-07-03 (uses 2022-07 data)
  - SatWet - no cached weights: 119.1s
  - SatWet - cached weights: 96.4s
  - This branch - no cached weights: 42.0s
  - This branch - cached weights: 19.0s

Performance in this branch is significantly better, with both cached and uncached scenarios completing 3x quicker. Computing climatology doesn't have a significant impact.


aethr:
Merging this before we run a test run in sandbox, as it improves performance and speeds up tests. I've added it to the list of PRs for Peter to review, and will make any suggested changes in a follow up PR.

Issue #178, opened 2026-04-28
Handling updates to direct data sources


aethr:
## The problem

The prior has made an intentional move towards "direct" data sources, where data is fetched from primary data sources "on demand" when the prior sector which uses them needs them. Previously, many datasets were sourced from primary data sources, sometimes reformatted, and then made available via the public data store in S3.

Fetching directly from the primary source has a number of advantages:
- auditability and clarity of data provenance
- transformation and filtering of the data are documented as code
- updates to datasets don't require manual labour

However, this strategy does introduce additional challenges and considerations:
- downtime of a single dataset provider can halt the entire process
- updates to datasets that change field structure or values can break integrations
- updates to datasets can result in significant changes to results

### Unexpected updates

For static datasets or datasets that cover a large time period (ie, annual), unexpected updates to the dataset might not be desired from a scientific perspective. For example, recent updates to the ANGA inventories included significant changes to historical emission estimates in several sectors. If we were in the middle of a month of daily runs when this update arrived, the daily results on either side of this update would have had significant disparities across several sectors.

Although results after a dataset update may be more accurate, the sudden change in results might be misinterpreted as an actual change in events "on the ground" unless the dataset update is investigated and understood.

## Proposed solution

If we can delineate between dynamic and static data sources, it should be feasible to fetch all static data sources ahead of a planned batch of prior runs, and share static inputs across multiple prior invocations. This solves all three challenges listed above:
- once static datasets are fetched and archived, only public datastore (S3) must be available during prior processing
- structural changes to datasets only need to be resolved in the lead-up to data fetching runs
- all prior runs within a batch are guaranteed to operate on the same static inputs

To implement this solution, we would need:
- a mechanism for annotating datasets as "static", and annotation of all existing DataSources
- an infrastructure mechanism to share static inputs across prior runs within a "batch" or period

Both of these mechanisms would require careful thought in their design.

### Archiving shared inputs

Our current practice is to archive the entire working folder for a daily run in S3 on completion, allowing us to accurately reconstruct the inputs to a workflow run for later analysis. However, when the prior uses the same static inputs run after run, these end up duplicated in our object storage.

One beneficial side-effect to sharing a single "static" input folder across multiple runs is the de-duplication of inputs when saved to the archive. However, how would we represent this in the archive? The shared inputs themselves would need to be archived, and the location saved in each daily archive somehow so it could be re-constructed.

## Alternative solution

[...]

PR #188, opened 2026-05-17
Simplify domain config

aethr:
## Description

Before refactoring the PriorConfig and DataManager to support static inputs in support of #178, it makes sense to simplify how the domain is handled in the config.

The most substantial change is moving the inventory domain out of the config and into the inventory data module, and removing the env var used to configure it. The inventory domain should not be configurable, as its purpose is to describe the geographic extent of the emissions captured in the national inventory. We've created an `inventory_mask` in the `aust10km` domain which accomplishes this, and providing an alternative inventory domain would be incorrect without also changing the inventory data source.

While the domain remains in PriorConfig, fetching it has been simplified, and parsing has been delegated to a new `Domain` class. A `Domain` represents a dataset with its accompanying Grid definition. Both the domain and inventory domain can utilise this concept since they are backed by the same format on disk.

## Checklist

- [ ] Tests added


aethr:
As discussed, I'm merging this without approval so that we can get the majority of new features into `main` and run a full prior. I'll incorporate any feedback from review into a follow-up PR.

PR #190, opened 2026-05-19
Introduce separated PriorConfig and PriorParameters

aethr:
## Description

PriorConfig has grown over time to incorporate more and more functionality, providing access to filesystem paths, domain Grid and Dataset, and prep functions. It is available in almost all sector methods, and a large number of helpers. Dependency on this large class makes unit testing difficult. To test a function that works on dates, a domain file must be present, etc.

This PR simplifies PriorConfig to provide mostly static system configuration, plain config values read from the environment. It is no longer passed down to DataSource or Sector implementations which don't deal with files or runtime directly.

PriorParameters has been introduced to contain only the run-time parameters that are likely to change from run to run, and which control the domain and period over which each sector must compute its emissions. Data management, like fetching and loading files and intermediate storage is now provided by the DataManager.

These changes introduce a smaller "configuration surface" for Sectors and DataSources to operate within. It should be possible to create helper functions like "how many seconds are in the period of interest" without having to construct PriorConfig objects with inventory domains and file paths.

## Checklist

- [ ] Tests added


aethr:
At this point, I think this change is to large, for too little concrete benefit. I'll leave the branch for the time being, but I'm closing this branch so it doesn't clutter our workspace.

PR #193, opened 2026-05-22
Store static data in a separate path

### LINKED ISSUE #178: Handling updates to direct data sources

aethr:
## Description

Resolves #178.

This PR adds a new `static_path` PriorConfig option, and an optional `dynamic` flag to DataSource definitions. When a `static_path` is provided, PriorConfig and DataManager will treat the `input_path` as a storage location for dynamic inputs, and `static_path` as a storage location for static (non-dynamic) inputs.

A data source is considered "dynamic" if it may need to be fetched from run to run, or based on parameters such as the period or domain of interest. GFAS is an example of a dynamic data source, as GFAS data is fetched for the specific dates being modelled.

From a software design perspective, it would probably be safer to treat all data sources as dynamic unless otherwise specified, because this would guarantee the data source is fetched on every run. However, most data sources used by the prior are static, and this appears to be the norm in this type of scientific computing. Most data is provided as static files representing a snapshot of data at a point in time which can be interpolated, or files which represent large periods of time, ie yearly intervals. Even data sources which are fetched via API (see the oil and gas sector) are rarely updated, so treating them as static by default is safer than receiving a large update part-way through a run of dailies.

## Checklist

- [x] Tests added

## Notes

One small addition to this PR is writing the PriorConfig into a file in the output folder on success. This is done primarily for reproducability, to make it easier to re-create the conditions that this prior was run under.

This change is included in this PR is because previously all production run inputs were captured in an archive, but moving forward the contents of the STATIC_INPUTS path will sit outside the run folder and so won't be included in the archive. By recording the PriorConfig `static_path` in this file, we'll make it easier to locate the input batch which was used in the prior run.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -9,6 +9,7 @@ DOMAIN_FILE=https://openmethane.s3.amazonaws.com/domains/au-test/v1/domain.au-te
 #INPUTS=data/inputs
 #OUTPUTS=data/outputs
 #INTERMEDIATES=data/intermediates
+#STATIC_INPUTS=data/static
 
 # Contents of this folder will be copied into the INPUTS folder before each
 # run, and INPUTS will be cached here after each run. Can reduce refetches of
--- a/.gitignore
+++ b/.gitignore
@@ -213,9 +213,7 @@ pyrightconfig.json
 
 # Custom ignores
 
-outputs
-inputs
-intermediates
-processed
+# Default data paths
+data/*
 
 !src/openmethane_prior/lib
--- a/changelog/193.feature.md
+++ b/changelog/193.feature.md
@@ -1,0 +1,1 @@
+Allow storing static data in a separate path
--- a/scripts/omPrior.py
+++ b/scripts/omPrior.py
@@ -54,3 +54,7 @@ if __name__ == "__main__":
     # write the output to file
     config.output_file.parent.mkdir(parents=True, exist_ok=True)
     prior_ds.to_netcdf(config.output_file)
+
+    # write config into output folder on success
+    with open(config.output_path / "config.yaml", "w") as config_out:
+        config_out.write(config.to_yaml())
--- a/src/openmethane_prior/lib/config.py
+++ b/src/openmethane_prior/lib/config.py
@@ -51,6 +51,9 @@ class PriorConfig:
     """Filename to write the prior output to as a NetCDFv4 file in
     `output_path`"""
 
+    static_path: pathlib.Path = None
+    """Filesystem path where static input files exist or should be fetched to."""
+
     input_cache: pathlib.Path = None
     """If provided, a local path where remote inputs can be cached."""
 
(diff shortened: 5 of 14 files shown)


aethr:
As discussed, I'm merging this without approval so that we can get the majority of new features into main and run a full prior. I'll incorporate any feedback from review into a follow-up PR.

PR #194, opened 2026-05-27
UNFCCC codes from unofficial ANGA API

aethr:
## Description

Recently we discovered that the recent publication of the FY 2023-24 ANGA inventory included some changes to level names of some sectors, and removal of some previously available sectors. This led to our manually created UNFCCC codes CSV to be out of date in a way that was hard to detect.

Instead of manually cross-referencing and fixing the UNFCCC codes CSV, we looked at how the inventory sectors are displayed on the ANGA website where the level names match exactly the levels in the inventory dataset. The website uses an undocumented, unofficial API which includes the "Paris inventory" sector hierarchy used to select sectors in the interactive inventory screen.

By fetching and parsing this API, we can build a list of UNFCCC codes which exactly match the level names present in the ANGA inventory. This only needs to be generated once and then it can be added to the public data store. Alternately, assuming the unofficial API remains available, fetching the codes at the same time the inventory is fetched should guarantee that level names in both datasets always match.

## Checklist

- [x] Tests added

### DIFF
--- a/changelog/194.improvement.md
+++ b/changelog/194.improvement.md
@@ -1,0 +1,1 @@
+Fetch UNFCCC codes and level names from unofficial ANGA website API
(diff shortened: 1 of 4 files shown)


aethr:
Merging this without review, as it's a critical fix to ensure we don't miss any inventory emissions. I've added this PR to Peter's list of PRs to review, and I'll make any suggested changes in a follow up PR.

PR #195, opened 2026-05-28
Fix `fetch_domain` not working with file paths

aethr:
## Description

The DOMAIN_PATH env variable must work with URLs or local paths, but in it's current iteration local paths are not supported. This change checks if the provided domain is a file path, and copies it (via hardlink) into the correct location if so.

The docs for urllib's urlretrieve state that the function works with local file paths, but I misunderstood this to mean that a file path could be supplied as the url parameter, which is not the case. My local testing didn't catch this because I was using INPUT_CACHE locally, which was copying the domain file into the inputs folder ahead of each test run.

## Checklist

- [x] Tests added

### DIFF
--- a/changelog/195.fix.md
+++ b/changelog/195.fix.md
@@ -1,0 +1,1 @@
+Fix prior failing when DOMAIN_PATH is a file path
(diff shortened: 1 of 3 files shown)


aethr:
Merging this without review after local testing, as it will unblock sandbox runs we want to run today.

PR #183, opened 2026-05-06
Update livestock sector implementation to use headcounts dataset

aethr:
## Description

Resolves #182.

Integrates the refreshed CSIRO livestock headcount dataset, and calculates CH4 emissions per animal based on static annual emission estimates. Previously this sector was only concerned with enteric fermentation, but since the two main sources of emissions from cattle and sheep are enteric fermentation and manure, this PR also moves manure management emissions into the livestock layer, and out of the agriculture layer, for cattle and sheep.

## Checklist

Please confirm that this pull request has done the following:

- [x] Tests added
- [ ] Documentation added (where applicable)
- [x] Changelog item added to `changelog/`

## Notes

Even after introducing a completely new data source and methodology, the livestock emissions haven't significantly changed, as evidenced by the minimal changes to the livestock layer in test_om_prior.py. However, since we're now spatialising the inventory total directly, we're seeing 0.0% discrepancy in the livestock layer when running the prior over the full domain:
```
verification:PASSED - livestock emissions OK, discrepancy is 0.0% of total
```
Previously, this sector was failing with between 0.3 - 0.8% discrepancy depending on the period of interest.

It's worth noting, the livestock headcounts dataset includes beef and dairy cows and sheep, but does not include swine or other livestock. Based on this we've moved the UNFCCC categories for enteric fermentation from swine and other livestock into the agriculture sector, which is spatialised by land use. This causes a small increase in the agriculture emissions because their combined sector total will be increased by this inclusion.

## ALUM codes

Now that the livestock layer accounts for all emissions from cattle and sheep, the ALUM categories included in the agriculture sector have been updated to remove:
- 2.1.0 Grazing native vegetation
- 3.2.0 Grazing modified pastures
- 4.2.0 Grazing irrigated modified pastures
- 5.2.1 Dairy sheds and yards
- 5.2.2 Feedlots

The visibility of these sector/category relationships was previously hard to find (located in `landuse-sector-map.csv` data file), so ALUM codes for the agriculture sector have been moved into the sector implementation in `sector.py`. For consistency, the same change has been made to the LULUCF sector, and the `landuse-sector-map.csv` data source has been removed from the project.

PR #196, opened 2026-05-28
Hard code sector ALUM codes

aethr:
## Description

This removes the landuse-sector-map.csv file where prior sectors are associated with ALUM codes, which is how the NLUM Land Use GeoTIFF designates land use of each pixel. The landuse-sector-map.csv was already tightly coupled to the sector implementation, because one of the columns was the sector name.

With this change we lose the ability to update which ALUM codes fall into each sector without updating the implementation. However, like UNFCCC and ANZSIC codes, the sector implementation is intimately bound to the types of activities and emissions it represents. It doesn't make sense to change the ALUM mapping for a sector without changing the implementation.

This also makes the ALUM mapping more visible, so that if the implementation does change (for example, the recent change to include swine and other livestock in the agriculture sector), it forces the person making the change to evaluate whether the mapped ALUM codes are still accurate.

## Checklist

- [ ] Tests added


aethr:
Discussed this with @prayner and decided to incorporate this into #183. It will make the shifting of UNFCCC codes and the corresponding relationship with ALUM codes more clear in that changeset.

PR #197, opened 2026-05-28
Integrate Safeguard Mechanism into Waste sector

aethr:
## Description

This takes a similar approach to Safeguard integration in the oil and gas sector. Possible waste sector emission sources are gathered from Climate TRACE and NPI facilities datasets. Safeguard Facilities are correlated with emission sources, where the reported facility emissions supersede the emission source estimate.

Once Safeguard facility emissions have been allocated, the remaining national inventory is spread across the remaining emission sources, scaled by the CT estimated `emissions_quantity`. NPI facilities don't have `emissions_quantity`, so they are assigned the average from the CT datasets.

## Checklist

- [x] Tests added


prayner:
oil_gas/sector.py:91: Perhaps add number of facilities found to
the diagnostic.
waste/sector.py:73: I'm not fluent in pandas but does the .loc on the
top-line there mean the top and bottom lines aren't consistent?

prayner (approved):

aethr:
> oil_gas/sector.py:91: Perhaps add number of facilities found to the diagnostic.

Sounds good.

> waste/sector.py:73: I'm not fluent in pandas but does the .loc on the
> top-line there mean the top and bottom lines aren't consistent?

Looking at waste/sector.py:73 in the updated file I don't see `.loc`, but it's used in line 73 before the changeset. This is the updated block for that section:
```python
    emission_sources_df.loc[unallocated_emission_sources_mask, "inventory_quantity"] = (
        unallocated_emissions_scale * emission_sources_df.loc[unallocated_emission_sources_mask, "emissions_quantity"]
    )
```

In pandas, `DataFrame.loc[index]` can be used to refer to a single row if `index` is a scalar value, or it can refer to multiple rows if `index` is a pandas `Index` or a mask. In these files I'm frequently constructing masks for the rows I want to update, and then writing to them like `df.loc[some_mask, "column_name"] = some_value`.

In these lines `some_value` could be a scalar value, or it could be a series, but if a series is supplied it has to match the length of `some_mask`, or it needs to share an index with the original DataFrame. In this particular block we have two columns: "emissions_quantity" which is an emissions estimate from ClimateTRACE, which will not be present in NPI emission sources, and "inventory_quantity" (which we're creating) which means "the amount of emissions from this source according to the inventory".

So this particular block is the one which scales ClimateTRACE emission estimates based on the relative proportion between the ClimateTRACE sum and the inventory sum for all sources, and we're only using this value for the mask of sources which don't already have a number assigned to "inventory_quantity".

PR #207, opened 2026-07-15
Remove NT well filtering by `PURPOSE` which is no longer available

aethr:
## Description

A recent update to the NT Petroleum Wells dataset has replaced the PURPOSE for every well with "NR", making it useless for us to filter by.

The [metadata](https://www.ntlis.nt.gov.au/metadata/export_data?type=html&metadata_id=2DBCB771210D06B6E040CD9B0F274EFE) for this dataset still lists the Purpose field with the values we previously saw and used for filtering, but the data is missing from the actual dataset.

## Checklist

- [x] Tests added

## Notes

I've raised an enquiry with the NT geoscience email address to understand whether this change was intentional. If this data is intentionally removed/hidden then this PR will have to be merged as a permanent change. If it turns out that the data was removed accidentally / temporarily, we should probably merge this in the short term so that the prior continues to function, but plan to revert the change later.

### DIFF
--- a/changelog/207.fix.md
+++ b/changelog/207.fix.md
@@ -1,0 +1,1 @@
+Remove NT oil well filtering based on the PURPOSE field which no longer contains accurate data
--- a/src/openmethane_prior/sectors/oil_gas/emission_sources/nt_sources.py
+++ b/src/openmethane_prior/sectors/oil_gas/emission_sources/nt_sources.py
@@ -35,10 +35,6 @@ def nt_emission_sources(
     nt_wells_df: gpd.GeoDataFrame = nt_wells_da.data
     nt_titles_df: gpd.GeoDataFrame = nt_titles_da.data
 
-    # filter out non-production wells
-    emitting_well_purposes = ["Development", "Production"]
-    nt_wells_df = nt_wells_df[nt_wells_df["PURPOSE"].isin(emitting_well_purposes)]
-
     # well datasets may have duplicate rows for a single location due to
     # further drilling at a site to deepen/extend an existing hole
     nt_wells_df = nt_wells_df.sort_values(by="DT_RELEASE")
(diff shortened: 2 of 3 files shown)


aethr:
Email to geoscience.spatial@nt.gov.au:

> I'm writing to inquire about a recent change to the Petroleum Wells dataset published via STRIKE. 
> 
> The metadata description for this dataset (https://www.ntlis.nt.gov.au/metadata/export_data?type=html&metadata_id=2DBCB771210D06B6E040CD9B0F274EFE) lists the PURPOSE field as:
> 
> > Purpose: Indicates the major purpose for the well as it was initially designed. (Appraisal, Development, Exploration, Extension-Appraisal, Stratigraphic)
> 
> Previously, we had seen all the listed values (Appraisal, Development, etc.) in the PURPOSE field of the Shapefile of this dataset. However, a recent update to the file seems to have removed these values so that every listed well now has a PURPOSE value of "NR". While this field is of particular interest to us, there may be other fields which have had their data replaced with "NR" in the most recent published dataset.
> 
> Is this change intentional, or is the missing data a temporary issue which may be resolved in the future?
> 
> If removing this data was intentional, what is the reasoning behind the change in data availability?

aethr:
Merging this without review, as it's currently breaking for anyone who tries to run the prior.

aethr:
I received a response from Northern Territory Department of Mining and Energy, it appears the change was not intentional and will be corrected soon:

> Thanks for informing us about this issue. We use FME to join data from database and an excel files. Recently we switched the Data Storage and Backup System. The reader from FME should adapt it automatically, but it didn’t. I reimported them and it will go to production as soon as possible.
> 
> Cheers!

aethr:
The restored data is now visible again in the production Petroleum Wells Shapefile. This PR will be reverted in #209.

